PR #14, opened 2020-02-01
Creates service-client lib

carsonfarmer:
Still trying to get the docker compose stuff to run two daemons for the tests, but otherwise, this is gtg.


andrewxhill:
wow, nice work! this looks like it was a bunch to figure out. i think the only thing that popped out at me was:

it's obvious that we use the types of the thread_service grpc api via the protobufs. but then otherwise, we recreate types that maybe should be singularly defined for all implementations of threads. so for example, the event type is defined as typescript here, why not import it from one of the pb libraries so we can ensure only one event type exists across all threads? maybe it's not a big deal, it just struck me as a good long term safety/upgrade measure.

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+      - master
+
+jobs:
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'skip-ci')"
+
+    services:
```
carsonfarmer: This doesn't really quite work. The two images are accessible via their respective API ports (5007 and 5008), but they _aren't_ visible to each other for networking, as the currently run on the same ports (4006).

packages/core/src/ipld.ts
```diff
@@ -0,0 +1,43 @@
+import CID from 'cids'
```
carsonfarmer: This is here because there are no types for @ipld/block, but its also a moving target at the moment, so not worth publishing the types to definitely typed or similar.

packages/encoding/src/record.ts
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * CID of previous record.
+   */
+  prev?: CID
+}
+
+/**
+ * EncodedRecord is a serialized version of a record that contains link data.
+ */
+export interface EncodedRecord {
```
carsonfarmer: _Could_ use the actual protobuf object here from service_proto, but since this is the _only_ interface we need from there, I haven't included that as a dependency. Makes this ending module slightly less dependent on our existing threads implementation per se...

packages/encoding/src/record.ts
```diff
@@ -0,0 +1,130 @@
...
+  const hnode = Block.decoder<Uint8Array>(rawHeader, opts.codec, opts.algo)
+  const rawBody = Buffer.from(proto.bodynode as string, 'base64')
+  const body = Block.decoder<Uint8Array>(rawBody, opts.codec, opts.algo)
+  const decoded = decodeBlock(rnode, keyiv)
+  const robj = decoded.decode()
+  const eobj = enode.decode()
+  enode.encode() // Created encoded value
+  body.encode()
+  hnode.encode()
+  const evt: Event = {
```
carsonfarmer: Event and LogRecord here as to be consumed via Javascript, and mimic the interfaces used in go-threads/core. They should probably be moved to js-threads/core as well.

packages/service-client/src/index.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+import { ThreadID, Variant, ThreadInfo, ThreadProtocol, Block } from '@textile/threads-core'
+import { createEvent, createRecord } from '@textile/threads-encoding'
+import Multiaddr from 'multiaddr'
+import { Client, RecordInfo } from '.'
+
+const hostAddrPort = 4006
+const proxyAddr = 'http://127.0.0.1:5007'
+const ed25519 = keys.supportedKeys.ed25519
+
+const protocol = {
```
carsonfarmer: This protocol stuff is merely ghost code for now. Since JS doesn't support custom protocols very nicely, we might have to rethink what we accept on the Go side here?

packages/service-client/src/index.ts
```diff
@@ -0,0 +1,276 @@
...
+import Multiaddr from 'multiaddr'
+import PeerId from 'peer-id'
+import { keys } from 'libp2p-crypto'
+import { ThreadID, ThreadInfo, KeyOptions, LogInfo, Block } from '@textile/threads-core'
+import * as pb from '@textile/threads-service-grpc/api_pb'
+import { API } from '@textile/threads-service-grpc/api_pb_service'
+import * as encoding from '@textile/threads-encoding'
+import { RecordNode } from '@textile/threads-encoding'
+
+export interface RecordInfo {
```
carsonfarmer: Should probably be moved to `js-threads/core`

packages/service-client/src/index.ts
```diff
@@ -0,0 +1,276 @@
...
+
+function threadInfoFromProto(proto: pb.ThreadInfoReply.AsObject) {
+  const id = ThreadID.fromBytes(Buffer.from(proto.id as string, 'base64'))
+  const readKey = Buffer.from(proto.readkey as string, 'base64')
+  const replicatorKey = Buffer.from(proto.followkey as string, 'base64')
+  const logs: Set<LogInfo> = new Set()
+  for (const log of proto.logsList) {
+    const rawId = Buffer.from(log.id as string, 'base64')
+    const pid = PeerId.createFromBytes(rawId)
+    // @todo: Currently it looks like private key unmarshaling isn't compatible between Go and JS?
```
carsonfarmer: JS expects a different format from what we're getting back from Go. Need to double check, but it looks like it is expecting the public key to be appended to the end of the buffer...?

packages/service-client/src/index.ts
```diff
@@ -0,0 +1,276 @@
...
+    logs,
+  }
+  return threadInfo
+}
+
+/**
+ * Client is a web-gRPC wrapper client for communicating with a webgRPC-enabled Textile server.
+ * This client library can be used to interact with a local or remote Textile gRPC-service.
+ */
+export class Client {
```
carsonfarmer: Implements the full thread service API via the Go gRPC API

packages/service-client/src/index.ts
```diff
@@ -0,0 +1,276 @@
...
+      return cb(
+        {
+          record,
+          threadID: id,
+          logID,
+        },
+        err,
+      )
+    }
+    return grpc.invoke(API.Subscribe, {
```
carsonfarmer: Returns Response with `cancel` function. Note that this is actually the correct way to handle server-side streams, and `js-threads-client` should be updated to this pattern,

carsonfarmer:
I think I'm close with the CI test docker setup. I can't quite figure out how to have the two images able to connect to one another. I think I'm incorrectly specifying their local addresses (should be using docker names?), but don't want to invest much more time here given its probably an obvious fix.

.github/workflows/test.yml
carsonfarmer: To clarify, I'm use they _are_ visible, I'm just not sure _how_ to make them visible to each other.

andrewxhill:
```
      threads1:
        image: textile/go-threads
        ports:
          - 4006:4006
          - 5006:5006
          - 5007:5007
          - 6006:6006
          - 6007:6007
      threads2:
        image: textile/go-threads
        ports:
          - 4106:4006
          - 5106:5006
          - 5107:5007
          - 6106:6006
          - 6107:6007
```

andrewxhill: or to be extra safe, maybe something like this would be pretty redundant,

```
      threads1:
        image: textile/go-threads
        ports:
          - 127.0.0.1:4006:4006
          - 127.0.0.1:5006:5006
          - 127.0.0.1:5007:5007
          - 127.0.0.1:6006:6006
          - 127.0.0.1:6007:6007
      threads2:
        image: textile/go-threads
        ports:
          - 127.0.0.1:4106:4006
          - 127.0.0.1:5106:5006
          - 127.0.0.1:5107:5007
          - 127.0.0.1:6106:6006
          - 127.0.0.1:6107:6007
```

packages/service-client/src/index.ts
carsonfarmer: done

carsonfarmer:
This is all GTG now @andrewxhill, tests passing now thanks to updated go-threads image.

packages/encoding/src/record.ts
andrewxhill: makes sense. i think moving in that direction over time will be good though, so any changes to the ref implementation will near-immediately (thanks dependabot) show as breaking in this lib

PR #15, opened 2020-02-05
Higher-level Service API with local key handling

carsonfarmer:
This sub-package is a light-weight wrapper around service-client that stores (caches) local key information. This makes it easier for consuming applications to store key information required for local encryption/encoding, without sending any private information to the remote threads service. Since it takes a Client as argument, it should play nicely with Textile's remote Threads services etc.

The tests are copy-pasta from service-client, which is great, because it means this lib provides the _same_ behaviour as service-client, even when all keys are kept with the user/locally.


carsonfarmer:
Seeing some timeout errors on this one... not sure where those are coming from... Will have to dig in later tonight.

carsonfarmer:
Ok, re-ran and its going fine @andrewxhill.

PR #17, opened 2020-02-06
Custom multiaddr implementation that supports our thread protocol

carsonfarmer:
Unfortunately, js-multiaddr does not easily support custom protocols without essentially updating the hard-coded protocols table (and convert.js file) or forking the library. This option (creating our own _internal_ variant), seems the lesser of several possible evils? There is an ongoing discussion about a better way forward to support custom protocols (https://github.com/multiformats/js-multiaddr/issues/22), which we should track so that we can avoid this ugliness in the future...


carsonfarmer:
🤠

PR #21, opened 2020-02-11
Chat example + various edge case fixes and API improvements

### LINKED ISSUE #20: Port chat example to JS
Port (some variant of) https://github.com/textileio/go-threads/tree/master/examples/chat to JS/TS

carsonfarmer:
## Description

This adds a chat example that is fully compatible with the go-threads chat example. It provides e2e encrypted chat, using a remote threads daemon for the service layer APIs. This should provide a useful example for external devs exploring the lower-level threads APIs.
Fixes #20 

## Type of change

## How Has This Been Tested?

For all relevant packages, tests have been run and updated as needed. They run locally using the same setup as our default CI. No tests were added to the example sub-package, as this _is_ essentially the test.

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint": "eslint --fix --ext .js,.ts packages/*/src/**/*",
     "test": "npm run test:node && npm run test:browser",
     "test:node": "TS_NODE_FILES=true mocha --exit",
-    "test:browser": "lerna exec 'polendina --cleanup --timeout 60 src/**/*.spec.ts --webpack-config ../../webpack.test.js'",
+    "test:browser": "lerna run test:browser",
     "docs": "typedoc && touch ./docs/.nojekyll"
   },
   "author": "Textile <contact@textile.io>",
--- a/packages/core/src/multiaddr/index.ts
+++ b/packages/core/src/multiaddr/index.ts
@@ -1,10 +1,8 @@
 import varint from 'varint'
-import multibase from 'multibase'
 import CID from 'cids'
+import bs58 from 'bs58'
 import { protocols } from './protocols'
 import * as codec from './codec'
-
-const bs58 = (buf: Buffer) => multibase.encode('base58btc', buf)
 
 /**
  * Creates a [multiaddr](https://github.com/multiformats/multiaddr) from
@@ -266,7 +264,7 @@ export class Multiaddr {
       // Get the last id
       b58str = (tuples.pop() || [])[1]
       // Get multihash, unwrap from CID if needed
-      b58str = bs58(new CID(b58str || '').multihash).toString()
+      b58str = bs58.encode(new CID(b58str || '').multihash)
     } catch (e) {
       b58str = undefined
     }
--- a/packages/core/src/thread/log.ts
+++ b/packages/core/src/thread/log.ts
@@ -14,7 +14,7 @@ export interface LogInfo {
   /**
    * The logs public key used to check signatures.
    */
-  pubKey: PublicKey
+  pubKey?: PublicKey
   /**
    * The logs private key, used to sign content when available.
    */
--- a/packages/encoding/src/index.spec.ts
+++ b/packages/encoding/src/index.spec.ts
@@ -36,7 +36,7 @@ describe('Encoding...', () => {
       const privKey = await keys.generateKeyPair('Ed25519', 256)
       const body = Block.encoder(raw, defaultOptions.codec)
       const event = await createEvent(body, readKey)
-      const { value } = await createRecord(event, privKey, undefined, replicatorKey)
+      const { value } = await createRecord(event, privKey, replicatorKey, undefined)
       const decoded = decodeBlock<RecordNode>(value, replicatorKey).decode()
       expect(decoded.prev).to.be.undefined
       expect(decoded).to.haveOwnProperty('block')
--- a/packages/encoding/src/record.ts
+++ b/packages/encoding/src/record.ts
@@ -1,7 +1,7 @@
 import CID from 'cids'
 import log from 'loglevel'
 import { Block, Event, EventNode, LogRecord, RecordNode } from '@textile/threads-core'
-import { randomBytes, PrivateKey } from 'libp2p-crypto'
+import { PrivateKey } from 'libp2p-crypto'
 import { Options, defaultOptions, encodeBlock, decodeBlock } from './coding'
 
 const logger = log.getLogger('encoding:record')
@@ -20,8 +20,8 @@ export interface EncodedRecord {
 export async function createRecord(
   data: Event,
   privKey: PrivateKey,
+  keyiv: Uint8Array,
   prev?: CID,
-  key?: Uint8Array,
   opts: Options = defaultOptions,
 ) {
   logger.debug('creating record')
@@ -38,7 +38,6 @@ export async function createRecord(
   // Don't include prev unless it is defined
   if (prev) obj.prev = prev
   const node = Block.encoder(obj, opts.codec, opts.algo)
-  const keyiv = key || randomBytes(44)
   const value = encodeBlock(node, keyiv, opts)
   // @todo: We don't support a dag here yet, but this is where we'd add this data to IPFS!
   const record: LogRecord = {
--- a/packages/examples/README.md
+++ b/packages/examples/README.md
@@ -1,0 +1,57 @@
+# Threads Examples _(threads-examples)_
+
+[![Made by Textile](https://img.shields.io/badge/made%20by-Textile-informational.svg?style=popout-square)](https://textile.io)
+[![Chat on Slack](https://img.shields.io/badge/slack-slack.textile.io-informational.svg?style=popout-square)](https://slack.textile.io)
+[![npm (scoped)](https://img.shields.io/npm/v/@textile/threads-examples.svg?style=popout-square)](https://www.npmjs.com/package/@textile/threads-examples)
+[![docs](https://img.shields.io/badge/docs-master-success.svg?style=popout-square)](https://textileio.github.io/js-threads)
+[![standard-readme compliant](https://img.shields.io/badge/standard--readme-OK-green.svg?style=flat-square)](https://github.com/RichardLitt/standard-readme)
+
+> A set of example apps using Textile's Threads Protocol
+
+Join us on our [public Slack channel](https://slack.textile.io/) for news, discussions, and status updates. [Check out our blog](https://blog.textile.io) for the latest posts and announcements.
+
+## Table of Contents
+
+- [Background](#background)
+- [Install](#install)
+- [Usage](#usage)
+- [API](#api)
+- [Maintainers](#maintainers)
+- [Contributing](#contributing)
+- [License](#license)
+
+## Background
+
+TODO
+
+## Install
+
+```bash
+// TODO
+```
+
+## Usage
+
+```typescript
+// TODO
+```
+
+## API
+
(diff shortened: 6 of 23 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "lint": "eslint --fix --ext .js,.ts packages/*/src/**/*",
     "test": "npm run test:node && npm run test:browser",
     "test:node": "TS_NODE_FILES=true mocha --exit",
-    "test:browser": "lerna exec 'polendina --cleanup --timeout 60 src/**/*.spec.ts --webpack-config ../../webpack.test.js'",
+    "test:browser": "lerna run test:browser",
```
carsonfarmer: Did this throughout... makes running tests much easier, and seemingly works more consistently.

packages/core/src/multiaddr/index.ts
```diff
@@ -266,7 +264,7 @@ export class Multiaddr {
       // Get the last id
       b58str = (tuples.pop() || [])[1]
       // Get multihash, unwrap from CID if needed
-      b58str = bs58(new CID(b58str || '').multihash).toString()
+      b58str = bs58.encode(new CID(b58str || '').multihash)
```
carsonfarmer: We don't actually want a multihash here, just base58 encoding. This fixes that.

packages/core/src/thread/log.ts
```diff
@@ -14,7 +14,7 @@ export interface LogInfo {
   /**
    * The logs public key used to check signatures.
    */
-  pubKey: PublicKey
+  pubKey?: PublicKey
```
carsonfarmer: Needs to be optional.

packages/encoding/src/index.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('Encoding...', () => {
       const privKey = await keys.generateKeyPair('Ed25519', 256)
       const body = Block.encoder(raw, defaultOptions.codec)
       const event = await createEvent(body, readKey)
-      const { value } = await createRecord(event, privKey, undefined, replicatorKey)
+      const { value } = await createRecord(event, privKey, replicatorKey, undefined)
```
carsonfarmer: Ordering changed here and throughout.

packages/encoding/src/record.ts
```diff
@@ -20,8 +20,8 @@ export interface EncodedRecord {
 export async function createRecord(
   data: Event,
   privKey: PrivateKey,
+  keyiv: Uint8Array,
```
carsonfarmer: This should never have been optional. Otherwise its gets eaten up in the function. Fixes that.

packages/examples/README.md
```diff
@@ -0,0 +1,57 @@
+# Threads Examples _(threads-examples)_
```
carsonfarmer: Examples package can actually be installed separately... Currently that's not all that useful, but in the future, it would be nice to actually be able to install and run various examples directly from the cmd line.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
+/* eslint-disable import/first */
+// Some hackery to get WebSocket in the global namespace on nodejs
+;(global as any).WebSocket = require('ws')
```
carsonfarmer: Same as in various tests.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+
+// Create primary interface
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+  prompt: chalk.green('> '),
+})
+
+// CLI program
+const program = yargs(process.argv.slice(2))
```
carsonfarmer: The default program has most of the same options as the go version.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+  .alias({ a: 'apiAddr' })
+  .describe({ apiAddr: 'Threads service API gRPC proxy bind multiaddress' }).argv
+
+// Setup
+if (!fs.existsSync(program.repo)) {
+  fs.mkdirSync(program.repo, { recursive: true })
+}
+const store = new LevelDatastore(program.repo)
+const client = new Client({ host: program.apiAddr })
+const service = new Service(store, client)
```
carsonfarmer: Uses our service sub-package to keep the keys etc local. Only runs in nodejs obviously.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+
+// Local peer/thread information
+let peer: PeerId
+const peerKey = new Key('chat-peer-id')
+let thread: ThreadID = new ThreadID(Buffer.alloc(0))
+
+// Chat program
+let chat: any
+const resetCmds = () => {
```
carsonfarmer: This resetCmds is a bit of a hack to 'reset' the help each time a sub-command is run. Its annoying, but was recommended in the docs for Yargs 🤷‍♂ .

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+  chat = yargs()
+    .reset()
+    .help(':help')
+    .scriptName('')
+    .version('0.0.1')
+    .command(':address', 'Show the host or active thread addresses.', {}, async function() {
```
carsonfarmer: Same set of commands that the go version has, each prefixed with a colon.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+    })
+    .command(':exit', 'Exit the active thread, or whole program if no thread active.', {}, function() {
+      if (thread.defined()) {
+        thread = new ThreadID(Buffer.alloc(0))
+        rl.setPrompt(green('> '))
+      } else rl.close()
+    })
+    .command(
+      ':replicator <addr>',
+      'Add a replicator at address to active thread.',
```
carsonfarmer: Adds a remote replicator to the chat.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+      },
+    )
+    .command(':enter <name>', 'Enter thread with name.', { name: { type: 'string', required: true } }, async function({
+      name,
+    }) {
+      const info = await store.get(new Key(`/names/${name}`))
+      thread = ThreadID.fromBytes(info)
+      rl.setPrompt(green(`${name}> `))
+    })
+    .command(':keys', `Show the active thread's keys.`, {}, async function() {
```
carsonfarmer: Prints the (locally available) keys for the thread/chat.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+    .onFinishCommand(async () => {
+      rl.prompt()
+      resetCmds()
+    })
+}
+resetCmds()
+
+// Main program
+async function main() {
+  const has = await store.has(peerKey)
```
carsonfarmer: This just creates essentially an ephemeral key pair to use as our local identity. Could be any external source of keys.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+  } else {
+    peer = await PeerId.create({ keyType: 'Ed25519' })
+  }
+  // Welcome
+  console.log(grey('Welcome to Threads!'))
+  console.log(grey('Your peer ID is: ') + green(peer.toB58String()))
+  console.log(grey('Enter ') + green(':help') + grey(' for options.'))
+  rl.prompt()
+
+  service.subscribe(async (rec?: ThreadRecord, err?: Error) => {
```
carsonfarmer: All thread updates come thru here, just like the go version.

packages/examples/src/chat/index.ts
```diff
@@ -0,0 +1,247 @@
...
+        ' ' +
+        pink(id.slice(-7)) +
+        ' ' +
+        green(body.txt),
+    )
+    rl.prompt()
+  })
+
+  // Setup
+  rl.on('line', function(line) {
```
carsonfarmer: main program, just process inputs line by line.

packages/service/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from './store'
 export * from './service'
+export { Client } from '@textile/threads-service-client'
```
carsonfarmer: This wasn't exported before, is useful to have.

packages/service/src/service.ts
```diff
@@ -81,7 +79,11 @@ export class Service implements Interface {
    * @param id The Thread ID.
    */
   async getThread(id: ThreadID) {
-    return this.client.getThread(id)
+    const info = await this.client.getThread(id)
```
carsonfarmer: Also check for locally cached keys etc.

packages/service/src/service.ts
```diff
@@ -191,9 +189,10 @@ export class Service implements Interface {
     return info
   }
 
-  async getOwnLog(id: ThreadID, create: true): Promise<LogInfo>
-  async getOwnLog(id: ThreadID, create: boolean): Promise<LogInfo | undefined> {
-    const info = await this.client.getThread(id)
+  async getOwnLog(id: ThreadID, create?: true): Promise<LogInfo>
```
carsonfarmer: Better type behaviour.

packages/service/src/store/index.ts
```diff
@@ -63,8 +63,8 @@ export class LogStore {
    * @param info Log information.
    */
   async addLog(id: ThreadID, info: LogInfo) {
-    await this.keys.addPubKey(id, info.id, info.pubKey)
-    info.privKey && (await this.keys.addPrivKey(id, info.id, info.privKey))
+    if (info.pubKey) await this.keys.addPubKey(id, info.id, info.pubKey)
```
carsonfarmer: These are both optional now.

packages/examples/src/chat/index.ts
andrewxhill: any idea what happens here in React Native? i had to do some hackery to get crypto stuff into global in RN and the pattern was a smidge different, curious if it matters here

packages/examples/src/chat/index.ts
andrewxhill: ah, maybe this answers my RN question above?

packages/examples/src/chat/index.ts
andrewxhill: ah okay, disregard. I understand the example here now. makes sense

packages/examples/src/chat/index.ts
carsonfarmer: Well, this only runs in node because of the cli deps we're using. But actually, this comments is kinda misplaced, because nothing here dictates that this will only run in node... oops.

packages/examples/src/chat/index.ts
carsonfarmer: Yeh I think diff situation... having said that, I'm very interested to dig into your React Native solution for crypto stuff, because this is something others will also likely run into if they try to use this in RN at some point?

andrewxhill: https://github.com/textileio/js-threads-client/#buffer-not-found here is the setup to make js-threads-client work in RN

andrewxhill: but perhaps that could be solved with a solution like yours, internal to the library

carsonfarmer: For sure I think we could internal shim Buffer and friends, but the crypto stuff might be pretty tricky. Worth digging into for sure tho.

Issue #3, opened 2019-12-19
JS Threads: Listeners/Database Emitters


carsonfarmer:

carsonfarmer:
Ready for review in #31

carsonfarmer:
Completed with above PR.

Issue #2, opened 2019-12-19
JS Event Store: Store Manager


carsonfarmer:

carsonfarmer:
Closing this one as 'wont fix' as this is currently not a priority for the js-threads implementation.

PR #25, opened 2020-03-11
Local-first Thread Store (not Database)

### LINKED ISSUE #1: JS Event Store: Dispatcher

carsonfarmer:
## Description

This a WIP local-first Threads Store implementation. It provides the required underlying infrastructure for a fully-working, local-first, Threads Store. This module is primarily the lower-level datastore compliant store. It is not (really) meant to be the public API.

Fixes #1 #2 #3 and others

## Things left todo

These are things that should be done in separate, subsequent PRs.

- Documentation updates (creation)
- Better use of existing proto definitions for internal methods
- A working CRDT event codec

## Type of change

## How Has This Been Tested?

New tests are being written to reflect the new functionality. This is an entirely new module, with pretty much total test coverage.

## Checklist:

- [ ] I have made corresponding changes to the documentation


packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,90 @@
...
+   * @param events The (variadic list of) events to dispatch.
+   */
+  async dispatch<T extends Event>(...events: Result<T>[]) {
+    await this.writeLock()
+    try {
+      if (this.child) {
+        logger.debug('persisting events')
+        const batch = this.child.batch()
+        for (const { key, value } of events) {
+          batch.put(key.instance(ulid()), CborEncoder.encode(value))
```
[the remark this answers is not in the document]
carsonfarmer: Note that the key is an 'instance' of whatever was passed to the dispatcher from the store. The instance value is just a ulid, which is essentially an ordered UUID, useful for storing. By making the instance part of the key the lid, we get sorting, but we can still do things like prefix search/queries if we want as well. Something to consider for other dispatcher implementations.

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,90 @@
...
+        }
+        await batch.commit()
+      }
+      logger.debug('dispatching')
+      await Promise.all([...this.reducers].map((reducer: Reducer<T>) => reducer.reduce(...events)))
```
carsonfarmer: Waits until all reducers are done before returning and unlocking.

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,90 @@
...
+  async dispatch<T extends Event>(...events: Result<T>[]) {
+    await this.writeLock()
+    try {
+      if (this.child) {
+        logger.debug('persisting events')
+        const batch = this.child.batch()
+        for (const { key, value } of events) {
+          batch.put(key.instance(ulid()), CborEncoder.encode(value))
+        }
+        await batch.commit()
```
carsonfarmer: Wait until all events are persisted before reducing... Note that, if an error is thrown here, it bubbles all the way up to the caller. In our db implementation (not in this PR), we can actually retry with exponential backoff until this step works, but we _don't_ want to fire off the reducers before then...

packages/store/src/dispatcher.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+      logger.debug('dispatching')
+      await Promise.all([...this.reducers].map((reducer: Reducer<T>) => reducer.reduce(...events)))
+    } finally {
+      this.unlock()
+    }
+  }
+  // @todo: Should we have a close/stop method/behavior?
```
carsonfarmer: Probably, just to 'turn off' dispatching maybe?

packages/store/src/stores/basic.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { expect } from 'chai'
+import { MemoryDatastore, Key } from 'interface-datastore'
+import { collect } from 'streaming-iterables'
+import { Dispatcher } from '../dispatcher'
+import { BasicStore } from './basic'
+
+describe('BasicStore', () => {
+  describe('interface-datastore', () => {
```
carsonfarmer: A Basic store (the default 'Store' implementation) is also datastore compliant! What fun!

packages/store/src/stores/basic.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      setup() {
+        return new BasicStore(new MemoryDatastore(), new Key('test'), new Dispatcher())
+      },
+      teardown() {
+        return
+      },
+    })
+  })
+
+  it('should put pretty fast', async () => {
```
carsonfarmer: Ok, bad name for this test... tho it should...

packages/store/src/stores/basic.ts
```diff
@@ -0,0 +1,14 @@
+import { Datastore, Key, MemoryDatastore } from 'interface-datastore'
+import { Dispatcher } from '../dispatcher'
+import { BasicCodec } from '../codec'
+import { Store } from './store'
+
+export class BasicStore<T = any> extends Store<T> {
```
carsonfarmer: That's it, that's all that's required. Feed in the codec and you're good to go! The only reason we have this here is because the base Store class is abstract to indicate that it should be subclassed and extended.

packages/store/src/stores/jsonpatch.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { expect } from 'chai'
+import { MemoryDatastore, Key } from 'interface-datastore'
+import { collect } from 'streaming-iterables'
+import uuid from 'uuid'
+import { Dispatcher } from '../dispatcher'
+import { JsonPatchStore } from './jsonpatch'
+
+describe('JsonPatchStore', () => {
+  describe('interface-datastore', () => {
```
carsonfarmer: Whaaat? The JsonPatch store is _also_ datastore compliant? Well I never!?

packages/store/src/stores/jsonpatch.ts
```diff
@@ -0,0 +1,83 @@
+import { Datastore, Result, Key, MemoryDatastore } from 'interface-datastore'
+import jsonpatch, { Operation } from 'fast-json-patch'
+import { Dispatcher, Event } from '../dispatcher'
+import { Codec } from '../codec'
+import { Store, safeGet } from './store'
+
+/**
+ * Entity is any object with an ID field.
+ */
+export interface Entity {
```
carsonfarmer: This probably doesn't belong here... maybe in threads-core?

packages/store/src/stores/jsonpatch.ts
```diff
@@ -0,0 +1,83 @@
...
+  ID: string
+  [others: string]: any
+}
+
+/**
+ * Op is a custom Store operation with a specific set of types.
```
carsonfarmer: Note that this is slightly different from how go-threads jsonpatch is done. Because we don't have separate 'actions' and 'events' in that sense (because we have generics instead). This needs to be further integration tested with the go db implementation to make sure I did this right.

packages/store/src/stores/jsonpatch.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+export interface Op<T extends Entity> {
+  type: Op.Type
+  entityID: string
+  patch?: Operation[] | T
+}
+
+export class JsonPatchCodec<T extends Entity> implements Codec<T, Op<T>> {
```
carsonfarmer: Here's a nice example of a 'custom' Codec. onDelete is pretty simple (side note: since json patch _has_ a delete operation, and you can operate on the root of a doc, we _could_ have actually just used that), just return undefined for the patch. onPut is slightly more complicated, but again, pretty simple, just compare with the old and diff. Note that onPut is called for each event independently. I went back on forth on that, as it could be useful (as has been mentioned by @jsign in past conversations) to do some things with all events at once). I ended up going with one at a time because a) its easy for implementers to reason about I think, and b) because if you really want, you can just squash the events beforehand into a single event anyway.

packages/store/src/stores/jsonpatch.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (old === undefined) {
+      patch = { type: Op.Type.Create, entityID, patch: value }
+    } else {
+      const ops = jsonpatch.compare(old, value)
+      // If no ops, old == new
+      patch = { type: Op.Type.Save, entityID, patch: ops.length > 0 ? ops : old }
+    }
+    return patch
+  }
+  async onReduce(store: Datastore<T>, ...events: Result<Event<Op<T>>>[]) {
```
carsonfarmer: This is slightly more complicated to look at than I would have liked for an external developer, but I imagine this is actually about as complicated as you're gonna get in most cases anyway. For instance, ideally a 'create' event (just a new put) would actually be a patch operation that 'replaces' the root of the doc, then the extra union type for maybe T maybe Op[] wouldn't be needed. But to match up with go, we do it like this. Seems fine at the end of the day.

packages/store/src/stores/jsonpatch.ts
```diff
@@ -0,0 +1,83 @@
...
+            ? (patch as T)
+            : jsonpatch.applyPatch(prev, patch as Operation[]).newDocument
+        batch.put(newKey, merged)
+      }
+    }
+    return batch.commit()
+  }
+}
+
+export class JsonPatchStore<T extends Entity> extends Store<T, Op<T>> {
```
carsonfarmer: Again, here's all that's needed to actually create the class. It might be better to just make Store a valid class and not bother with sub-classing the abstract store class, but that can be changed in a future PR if we need to.

carsonfarmer: Note that I've completely left out the CRDT store from this PR. There is a draft one in the works, but it is pretty flaky, so it'll have to come in the future PR.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+import { EncodingDatastore, Encoder, CborEncoder } from '../datastores/encoding'
+import { DomainDatastore } from '../datastores/domain'
+import { Lockable } from '../datastores/abstract/lockable'
+// eslint-disable-next-line import/no-cycle
+import { Codec } from '../codec'
+
+/**
+ * Events for Store's EventEmitter
+ */
+type Events<T> = {
```
carsonfarmer: These are the events that the store emits... not to be confused with thread events, any other other type of events. These are just classic js event emitter events.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
+import { Datastore, Key, Result, Batch, Query, MemoryDatastore } from 'interface-datastore'
+import lexInt from 'lexicographic-integer'
+import { EventEmitter } from 'tsee'
+import { Semaphore } from '../datastores/abstract/lockable'
+import { Reducer, Dispatcher, Event } from '../dispatcher'
+import { EncodingDatastore, Encoder, CborEncoder } from '../datastores/encoding'
+import { DomainDatastore } from '../datastores/domain'
+import { Lockable } from '../datastores/abstract/lockable'
+// eslint-disable-next-line import/no-cycle
```
carsonfarmer: This cycle is just for the interfaces, so if these were moved out elsewhere, that might help remove this linter issue.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+ */
+type Events<T> = {
+  open: () => void
+  close: () => void
+  events: (...events: Event<T>[]) => void
+  update: (...updates: Update<T>[]) => void
+  error: (err: Error) => void
+}
+
+export type Action<T> = () => Promise<T>
```
carsonfarmer: This is a slightly simplified version of the one in dispatcher. These should probably be unified in the future. For now, this is just a simple 'deferred' promise-generating function.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+
+export interface Update<T = any> {
```
carsonfarmer: This is for listeners, note the added optional event here. In go we don't have this, I stuck it here because we can have the generic event without any extra work... so why not?

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+  id: string
+  collection: string
+  type?: string | number
+  event?: T
+}
+
+const asUpdate = <T = any>(event: Result<Event<T>>): Update<T> => {
```
carsonfarmer: DRY

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+  const { value } = event
+  return {
+    id: value.id,
+    collection: value.collection,
+    event: value.patch,
+  }
+}
+
+export const safeGet = async <T = any>(store: Datastore<T>, key: Key) => {
```
carsonfarmer: This is just a helper function to retuned undefined if a given key is not found, otherwise, pass along the error. Again, DRY.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+      return event
+    }
+    this.patches.push(deferred)
+  }
+
+  put(key: Key, value: D) {
+    const deferred = async () => {
+      const event: Event = {
+        timestamp: Buffer.from(lexInt.pack(Date.now())),
```
carsonfarmer: To patch with go, we pack the unix timestamp into lexicographically sorted bytes.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+        id: key.toString().slice(1),
```
carsonfarmer: These slice(1) are annoying, but Key automatically adds the trailing / to keys for some reason (maybe in go as well?)..?

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+        patch: await this.store.codec.onPut(this.store, key, value),
+      }
+      return event
+    }
+    this.patches.push(deferred)
+  }
+
+  async commit() {
+    if (this.patches.length > 0) {
+      await this.store.dispatch(...this.patches)
```
carsonfarmer: See store.dispatch below

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+    }
+    this.patches.length = 0
```
carsonfarmer: Important in case this batch gets reused (which, I guess you could, but you shouldn't).

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+    return
+  }
+}
+
+export abstract class Store<D = any, A = D> extends EventEmitter<Events<A>>
```
carsonfarmer: Stores are their own thing, not to be confused with the old stores in go. They are datastores that are lockable, are a reducer, can support arbitrary value encoding, and are an event emitter. They are used extensively in threads-db (not part of this PR).

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+  implements Lockable, Datastore<D>, Reducer<Event<A>> {
+  public child: Datastore<D>
+  readonly semaphore: Semaphore
+  constructor(
+    child: Datastore<any>,
+    public codec: Codec<D, A>,
+    public prefix: Key = new Key(''),
+    public dispatcher: Dispatcher = new Dispatcher(child),
```
carsonfarmer: Makes testing easier to have these defaults... but also, kinda useful to make these 'work' on their own as well.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+    public encoder: Encoder<D, Buffer> = CborEncoder,
```
carsonfarmer: Haven't needed anything else so far, might not be the case once we get a working CRDT store...

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  /**
+   * Acquire a read lock on a given key.
+   * The datastore is only allowed to acquire a lock for keys it 'owns' (any decedents of its prefix key).
+   * @param key The key to lock for reading.
+   * @param timeout How long to wait to acquire the lock before rejecting the promise, in milliseconds.
+   * If timeout is not in range 0 <= timeout < Infinity, it will wait indefinitely.
+   */
+  readLock(key: Key, timeout?: number) {
```
carsonfarmer: We (re) implement locking here, rather than extend that Lockable class because multiple inheritance is ugly, and because interfaces!

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+  /**
+   * Release current lock.
+   * Must be called after an operation using a read/write lock is finished.
+   * @param key The key to unlock.
+   */
+  unlock(key: Key) {
+    return this.semaphore.unlock(key)
+  }
+
+  async dispatch(...actions: Action<Event<A>>[]) {
```
carsonfarmer: All stores have a dispatch method, which actually does the dispatching to the dispatcher. This _could_ have been done on the batch object, but it is useful to have the store here directly. Note that we emit the events _before_ we dispatch them, but we only emit the updates _after_ they are dispatched. So consumers can listen to either end of the process if they want.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+        const value = await event()
+        const key = this.prefix.child(new Key(value.id))
+        return { key, value }
+      }),
+    )
+    this.emit('events', ...events.map(event => event.value))
+    return this.dispatcher.dispatch(...events)
+  }
+
+  open() {
```
carsonfarmer: It's a datastore... so we have all these methods. Could have just extended the shim datastore, but again... multiple inheritance...

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  has(key: Key) {
+    return this.child.has(key)
+  }
+
+  get(key: Key) {
+    return this.child.get(key)
+  }
+
+  put(key: Key, value: D) {
```
carsonfarmer: All writes are done in batch, just like in the go version.

packages/store/src/stores/store.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  query(query: Query<D>) {
+    return this.child.query(query)
+  }
+
+  batch(): ActionBatch<D, A> {
+    return new ActionBatch(this)
+  }
+
+  async reduce(...events: Result<Event<A>>[]) {
```
carsonfarmer: All stores are reducers. In fact, they are the only thing that actually implements the reducer interface in this whole process. Like in (say) flux, they receive all events, but by default, only operate on events they are concerned with. This could be changed, and we could even _not_ send them events they are concerned with, but this seems more flex and in line with our white paper etc.

carsonfarmer: Note that the update is emitted _afte_ we've completed reducing.

typedoc.json
```diff
@@ -15,5 +15,6 @@
   "preserveConstEnums": true,
   "stripInternal": true,
   "module": "commonjs",
-  "mdHideSources": true
+  "mdHideSources": true,
+  "lernaExclude": ["threads-example-chat"]
```
carsonfarmer: Just to make docs output nicer to look at.

packages/store/src/stores/jsonpatch.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    let count = 0
+    store.on('events', () => count++)
+    store.on('update', () => {
+      expect(++count).to.equal(2)
+      store.close().then(() => {
+        done()
+      })
+    })
+
+    store.put(new Key('bar'), { ID: uuid(), hello: 'world' }).then(() => {
```
carsonfarmer: ID is required, anything else goes. Note that the JsonPatch store has no concept of schemas. That is a threads-db thing (not part of this PR), and so it is left to the db to control what makes it into the store.

PR #31, opened 2020-03-24
Threads DB

carsonfarmer:
## Description

This is the JS implementation of Threads DB. See #25 for underlying details and the components used here.

### Details

Here is a (rough draft) general overview diagram to help in reviewing this work. As you can see, it reflects the [original diagram from the white paper](https://raw.githubusercontent.com/textileio/papers/master/figures/Architecture.png) quite closely (with some additional implementation details added here for reference).

draft-components

Here's the general idea:

A Database is tied to a single Thread (with associated Thread ID). There is currently no DB Manager as in Go, because it is more likely in a lightweight browser app or similar that a single DB per application is being used. A manager would be easy to add if/when needed.
A Database has an Event Emitter (in the nodejs sense), and Listeners can subscribe to events using 'wildcard' syntax in a way that is similar to how it works in Go. For example, you can do something like:

```
const db = new Database()
const Collection1 = await db.newCollectionFromObject('Collection1', {
  ID: '',
  name: '',
  counter: 0,
})

// This will listen to any and all event types on Collection1
db.emitter.on('Collection1.**', update => {
  console.log(update)
})

const thing = new Collection1({ ID: 'id-i1', name: 'Textile1' })
await thing.save()
```

Like in Go, a Database contains Collections, each of which are defined by a json-schema.org schema. These schemas define the 'shape' of Collection Instances (currently named Entities, this will be updated to reflect the Go naming soon). Collections themselves are sub-classes of a Store from the `@textile/threads-store` package. Stores implement the Datastore interface, so they are also key-value stores. Stores are a generic interface, and Collections implement a JsonPatch Store. Being generic, they could be made to support other types (CRDT-driven documents for instance) in the future.

Stores are defined by a Codec and a Reducer (whose actual logic is defined in the Codec, like in Go). As such, Collections also have a Codec and a Reducer, though these are currently just a JsonPatch Codec by default/only. So a Collection is a Store with a set of APIs to make it match our Document Store interface from the white paper. For example, there are Collection- and Instance-level APIs to work with data:

```
const i1 = new Collection1({ ID: 'id-i1', name: 'Textile1' }) // This is not yet persisted
await i1.save() // Save changes

// Modify the `i1` instance
i1.name = 'Textile0'
await i1.save() // Save changes

// Modify it again
i1.name = 'Blah'
i1.counter = 33

// Save it from the Collection
await Collection1.save(i1)

// Delete it from the Collection
await Collection1.delete(i1.ID)
```

Plus a bunch more APIs you'd expect, like insert, findOne, has, etc.. It also supports Mongodb style search (`find`), which returns an AsyncIterator that consumers can use:

```
const Thing = new Collection<Info>('things', {}) // Anything goes schema
await Thing.insert(
  { ID: '', other: -1, thing: 'five' },
  { ID: '', other: 2, thing: 'two' },
  { ID: '', other: 3, thing: 'three' },
  { ID: '', other: 4, thing: 'four' },
)

const all = Thing.find({ $or: [{ other: { $gt: 1 } }, { thing: 'one' }] }, { sort: { other: -1 } })
for await (const { key, value } of all) {
    console.log(value)
}
```

Collections also support read and write Transactions, but these are pretty basic at the moment. They are essentially just lockable, key-based states that you can put the Collection into, so that anything else that wants to write to the Collection must `await` for the transaction to finish. They do not yet provide isolation, though they do provide atomic update semantics. The limitations are simple due to a) not being available on any js-datastore implementations, and b) not _really_ being super important in the single-threaded world of Javascript.

As in Go, any mutations on the Collections (which are essentially aggregate roots), are dispatched via the Database's Dispatcher. There is one Dispatcher per Database in practice, and all Collections receive all updates, so their reducer is responsible for taking appropriate action. The Dispatcher is then responsible for 1) persisting the event, and 2) calling the Collections' Reducer methods. Same as in Go. At this point, the entire process can work entirely 'offline'. This design supports 'offline first' applications that may not have connections to a remote Peer for networking.

Networking is supported via the Event Bus. This is where js-threads sticks to the white paper perhaps more explicitly than the Go implementation? The Event Bus has two core components, a network Watcher (for observing updates from the Network layer) and a persistent Queue (for pushing updates to the Network layer). This actually does work quite similarly to the Go ThreadAdapter. Essentially, the Event Bus subscribes to events on the Database's Thread from the Network layer (currently called the Service layer, soon to be updated to match Go), and dispatches them via the Dispatcher to the Collections. From there, the behavior is identical to a 'local' event.

Conversely, for events generated locally on the Collections, these are pushed onto the Event Bus's Queue by the corresponding Collection (_after_ a successful dispatch process), and they will attempt to send the update out via the Network layer. The queue is persistent in the case of the app being offline. It will attempt to send the updates, with exponential backoff in the case of failures. At the moment, it will make 5 attempts before giving up and moving on to the next event. After it has processed its Queue, it will try any skipped events again (and again). A possible todo would be to eventually give up entirely on an event and eject it from the Queue, but it seems more likely that all events will fail than a single event. The Event Queue will continue to process events and flush them to the Network layer as long as there are new events coming in. Upon app restart, the Queue will restart from the top, so that (say) a page refresh would potentially lead to re-connecting to the remote Peer and processing local events again.

All of the above is backed by a single (or multiple, depending on how a developer wants to use them) Datastore. By using the Datastore interface, we can support any backend that supports the abstract-leveldown interface (which includes leveldb, mongo, sqlite, memory, and many more). Other backends could be implemented by implementing the Datastore interface for them.

### Design considerations

In areas where the JS implementation differs from the Go one, it was either due to trying to stick with JS conventions, or in order to simply the _external_ developer experience. For instance, the Codec interface is _slightly_ different, in that it limits things to two mutations and a reducer function. The usage is quite similar, but I felt this made it easier to use JS conventions to potentially extend the Collections. Additionally, in Typescript, we have really great support for generics, so these were used to good effect where possible. There are some improvements that could be made there for sure, but I think having them makes things around Actions and Events a bit simpler in JS-land. But at the end of the day, if I had a choice between making it simpler for an external Javascript developer and sticking to the Go version, I went with simplicity for Javascript. There are instances where this might lead to less 'optimized' code, but hey, we're running a decentralized database in the browser here... things are gonna get crazy!

### Still todo:

- [ ] Documentation!
- [ ] Indexing (there's a package existing for that already)
- [ ] Better handling of external crypto keys
- [ ] Full offline first capabilities (the infrastructure is there, just needs tests)
- [ ] Full compatibility testing with Go (manual tests have been done, needs proper DB testing)
- [ ] Demos, examples, and connectors with external identities
- [ ] React/Vue integration examples/tests/demos

## Type of change

## How Has This Been Tested?

Most new functionality has associated tests. Further compatibility testing between Go and Typescript is required. Additionally, it would be useful to have tests for some of the minor features of the DB, in particular, more testing for adding a remote Thread/DB, more testing for deleting db/threads, and more persistence testing. Tickets will be created for all of these items, but should not hold up reviews/PRs.

## Checklist:

- [ ] I have made corresponding changes to the documentation


packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+  async has(id: string) {
+    return this.child.has(new Key(id))
+  }
+
+  /**
+   * Find all entities matching the query
+   * @param query Mongodb-style filter query.
+   * @param options Additional options to control query operation.
+   */
+  find(query?: FilterQuery<T>, options: FindOptions<T> = {}) {
```
carsonfarmer: This implements a full MongoDB-style query syntax. Its _reasonably_ efficient because we're working with AsyncIterables, so they're essentially 'lazy'. But it does end up visiting a lot of entities this way. We _try_ to push as much of the query down to the key-value store, but filters are sorting are done in JS.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+    // @todo: The mingo types are incorrect here... so we hack around those a bit.
+    const m: mingo.default.Query = new (mingo as any).Query(query || {})
+    const filters: Query.Filter<T>[] = [({ value }) => m.test(value)]
+    const orders: Query.Order<T>[] = []
+    if (options.sort) {
+      for (const [key, value] of Object.entries(options.sort)) {
+        orders.push(items =>
+          items.sort((a, b) => {
+            return cmp(dot(a.value, key), dot(b.value, key), value || 1)
```
carsonfarmer: Here we're reaching down into the object to sort them, sorts aren't all that nice because of this... indexing will help, and this same procedure of reaching into the object with `dot` will apply here.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+   */
+  count(query: FilterQuery<T>, options: FindOptions<T> = {}) {
+    return reduce((acc, _value) => acc + 1, 0, this.find(query, options))
+  }
+}
+
+/**
+ * Collection is a store of entities defined by a single schema.
+ */
+export class Collection<T extends Entity = any> extends ReadonlyCollection<T> {
```
carsonfarmer: Here's the full writable-Collection, this is where the hackery exists. The constructor returns a function, that is 'cast' as a Collection because has all the properties of the Collection prototype assigned to it. So, it _is_ a Collection and a function, but it does require some fiddling here. The caller will never know the difference.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+      if (!entity.ID) entity.ID = uuid()
+      if (!c.validator(entity) && c.validator.errors) {
+        throw new ValidationError(c.validator.errors)
+      }
+      return new Document(c, entity) as Document<T> & T
+    }
+    Object.setPrototypeOf(self, this.constructor.prototype)
+    Object.getOwnPropertyNames(this).forEach(p => {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      Object.defineProperty(self, p, Object.getOwnPropertyDescriptor(this, p)!)
```
carsonfarmer: This is safe because we already checked that it has this property... the type checker just isn't able to know this for sure because it is a runtime check.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+    })
+    return self as Collection<T>
+  }
+
+  /**
+   * Save (multiple) entities.
+   * @note Save is similar to insert, except it allows saving/overwriting existing entities.
+   * @param entities A variadic array of entities.
+   */
+  save(...entities: T[]) {
```
carsonfarmer: All mutations happen within a batch. This makes them atomic and pretty safe, but it also means we can intercept them and create out actions/events from here. See [`ActionBatch`](https://github.com/textileio/js-threads/blob/master/packages/store/src/stores/store.ts#L55) from the Store API for details on how these batches work. Quite similar to the Go variant, except here, we create essentially 'deferred' actions (promises that won't resolve until we tell the to) that are applied when the batch is commited.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  /**
+   * Insert (multiple) new entities.
+   * @note Insert is similar to save, except it will not allow saving/overwriting existing entities.
+   * @param entities
+   */
+  async insert(...entities: T[]) {
+    // By convention we'll use insert here, but could use a specific key instead
+    const lockKey = new Key('insert')
+    await this.child.readLock(lockKey)
```
carsonfarmer: We actually lock the db for inserts because we need to do a check as well as a mutation. For the other mutations, we don't need to perform this check, so they are already atomic. This is just an alternative usage of the writeTransaction setup (same internal workings, less function call overhead).

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+        if (await this.child.has(key)) {
+          throw existingKeyError
+        }
+        if (!this.validator(entity) && this.validator.errors) {
+          throw new ValidationError(this.validator.errors)
+        }
+        batch.put(key, entity)
+      }
+      return await batch.commit()
+    } finally {
```
carsonfarmer: The unlock happens in a finally block so that is _always_ unlocks, even in the event of a failed commit.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+
+  /**
+   * Lock the collection for readonly operations.
+   * @param cb A callback that takes a readonly collection.
+   * @param timeout How long to wait to obtain the read lock. If after timeout seconds the lock
+   * is not obtained, the method will error.
+   * @note Currenrly, transactions are not atomic, or even really transactions in the normal sense.
+   * They only provide locking for multi-read/single-write access to the given collection.
+   */
+  async readTransaction(cb: (c: ReadonlyCollection<T>) => Promise<void> | void, timeout?: number) {
```
carsonfarmer: These are here because, they're kinda nice to have. But I haven't found them all that useful TBH. They need more tests, but they work, they are potentially useful for keeping updates 'safe', and folk'll probably ask about it. Essentially, they just lock the Collection (and _only_ that collection) for read or write purposes. You can have multiple reads in flight, but only one write. And if you're writing, they no new reads until the write is done. Pretty classic RW Lock semantics.

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/**
+ * Runs a complex db use-case, and returns events received according to the provided options.
+ * @param los: The action path to listen on. Paths are structured as: <collection>.<id>.<type>,
+ * and support wildcards. For example, <collection>.*.* will listen to all events for a given
+ * collection. Double wildcard (**) is recursive, so <collection>.** will also listen to call
+ * events on a collection. Similarly, ** listens to all events on all collections.
+ */
+async function runListenersComplexUseCase(los: string[]) {
```
carsonfarmer: This is the same exact setup that is used in Go. Same 'listener' tests as well.

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+  const db = new Database() // Use the defaults
+  const Collection1 = await db.newCollectionFromObject<DummyEntity>('Collection1', {
+    ID: '',
+    name: '',
+    counter: 0,
+  })
+
+  const Collection2 = await db.newCollectionFromObject<DummyEntity>('Collection2', {
```
carsonfarmer: These steps provide a pretty nice view into the API and how it feels. Multiple ways to do something. I write like the API, it feels very JS-friendly, and is inline with our white paper and Mongodb more generally.

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+    counter: 0,
+  })
+
+  // Create some instance *before* any listener, just to test that it doesn't appear on a
+  // listener's "stream".
+  const i1 = new Collection1({ ID: 'id-i1', name: 'Textile1' })
+  await i1.save()
+  await delay(1000)
+
+  const events: Update[] = []
```
carsonfarmer: Here's the emitter stuff in practice. I have a question here about maybe just wrapping this up into a 'listen' method like in Go, might be preferable, though this felt a but more JS-friendly maybe?

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+  // Collection1 save i1
+  // Collection1 save i2
+  await Collection1.writeTransaction(async c => {
+    i1.counter = 30
+    i2.counter = 11
+    await c.save(i1, i2)
+  })
+
+  // Collection2 save j1
+  j1.counter = -1
```
carsonfarmer: All the normal mutation syntax is available. At first I didn't like this, but @sanderpick convinced me otherwise, and I'm glad he did. Feels really nice to just treat Instances like normal JS objects!

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+  // Save i2
+  // Collection2 Save j1
+  // Delete i1
+  // Collection2 Delete j1
+  // Delete i2
+  return events
+}
+
+describe('Database', () => {
+  describe.skip('end to end test', () => {
```
carsonfarmer: This is being skipped because it sometimes takes too long in CI. But it runs locally just fine, and is pretty awesome to see! I always skip it in the browser, so might turn it back on for node tests in CI if I think we need it. This test was a direct port from Go.

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+  describe('Persistence', () => {
+    const tmp = 'database.db'
+    after(() => {
+      level.destroy(tmp, () => {
+        return
+      })
+    })
+
+    it('should work with a persistent database and custom options', async function() {
+      if (isBrowser) return this.skip()
```
carsonfarmer: We skip this in the browser due to how out browser testing framework works. This isn't great, so I'd like to find a solution for this at some point. But it isn't a show-stopper either. It _runs_ in a browser, just not cleanly in a headless browser.

carsonfarmer: P.S. I wasted a lot of time trying to figure this out, so if anyone wants to dig in here... be my guest!

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+      })
+      // Re-do again to re-use id. If something wasn't closed correctly, would fail
+      await db.close() // Closes eventBus and datastore
+      await db.open(id)
+      expect(db.collections.size).to.equal(1)
+      await db.close()
+    })
+  })
+
+  describe('Events', () => {
```
carsonfarmer: Ported over from Go

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
+import toJsonSchema from 'to-json-schema'
```
carsonfarmer: Finally, the DB implementation !

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+const schemaKey = metaKey.child(new Key('schema'))
+const duplicateCollection = new Error('Duplicate collection')
+
+export type Options = {
+  dispatcher?: Dispatcher
+  eventBus?: EventBus
+  service?: Service
+}
+
+export class Database {
```
carsonfarmer: Here are all the components of the DB. Service needs renaming, but mostly otherwise we're almost a perfect match with Go and the white paper. See main PR description and diagram for a high-level explanation of all of these components.

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+  public child: DomainDatastore<any>
+
+  public emitter: EventEmitter2 = new EventEmitter2({ wildcard: true })
+
+  /**
+   * Database creates a new database using the provided thread.
+   * @param datastore The primary datastore, and is used to partition out stores as sub-domains.
+   * @param options A set of database options.
+   */
+  constructor(datastore: Datastore<any> = new MemoryDatastore<any>(), options: Options = {}) {
```
carsonfarmer: So, we have default behavior for all of these. That makes tests easy, but maybe we _shouldn't_ have such defaults in practice. Because right now, the default store is an in memory datastore... which isn't maybe what someone is expecting?

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * fromAddress creates a new database from a thread hosted by another peer.
+   * @param addr The address for the thread with which to connect.
+   * Should be of the form /ip4/<url/ip-address>/tcp/<port>/p2p/<peer-id>/thread/<thread-id>
+   * @param threadKey Set of symmetric keys.
+   * @param datastore The primary datastore, and is used to partition out stores as sub-domains.
+   * @param options A set of database options.
+   */
+  static async fromAddress(
```
carsonfarmer: This (and other methods) should _mostly_ match the Go versions. There are some differences, but I've tried to keep them really similar. If we want changes, I'd like to do those in a subsequent PR if possible, though serious omissions or changes should obviously happen here!

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+    addr: Multiaddr,
+    threadKey?: ThreadKey,
+    datastore: Datastore<any> = new MemoryDatastore<any>(),
+    options: Options = {},
+  ) {
+    const db = new Database(datastore, options)
+    const info = await db.service.addThread(addr, { threadKey })
+    await db.open(info.id)
+    await db.service.pullThread(info.id)
```
carsonfarmer: This works nicely in practice (to await this pull), though it _might_ be better to return right away and just let that pull call fly off on its own?

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+    const schema = toJsonSchema(data) as JSONSchema
+    return this.newCollection<T>(name, schema)
+  }
+
+  /**
+   * newCollection creates a new empty collection from an input JSON schema.
+   * @param name A name for the collection.
+   * @param schema A valid JSON schema object.
+   */
+  async newCollection<T extends Entity>(name: string, schema: JSONSchema) {
```
carsonfarmer: Pretty much the most important thing here I guess? Matches quite closely with the Go implementation. Creates the collections, adds it to the Map, listens to its events, and gives it back to the caller. In the future, it might be nice to provide a way to specify the codec to use for the Collection, but I don't know what that use case is yet?

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Open the database.
+   * Opens the underlying datastore if not already open, and enables the dispatcher and
+   * underlying services (event bus, network service, etc). If threadID is undefined, and the
+   * database was already bootstrapped on a thread, it will continue using that thread. In the
+   * opposite case, it will create a new thread. If threadID is provided, and the database was
+   * not bootstrapped on an existing thread, it will attempt to use the provided threadID,
+   * otherwise, a thread id mismatch error is thrown.
+   */
+  async open(threadID?: ThreadID) {
```
carsonfarmer: I wanted to avoid having to do this (and I'll still try to find a way), but this was what I ended up on. Essentially, because much of the setup for a Datastore is async, it doesn't work to do this all in the Database's constructor method. Additionally, it is useful to be able to create the Database _before_ actually starting anything like the event bus or network layer right away. For instance, you can specify schemas etc before calling open.

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+    }
+    await this.rehydrate()
+    await this.eventBus.start(this.threadID)
+    this.eventBus.on('record', this.onRecord.bind(this))
+  }
+
+  /**
+   * Return db (remote) address and keys.
+   */
+  async getInfo() {
```
carsonfarmer: This is kinda lame, because we don't actually know much about the addresses of our remote Peer here. But since it probably isn't going to be used at all anyway, I haven't invested more time here. It should be improved.

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Close and stop the database.
+   * Stops the underlying datastore if not already stopped, and disables the dispatcher and
+   * underlying services (event bus, network service, etc.)
+   */
+  async close() {
+    this.collections.clear()
+    await this.eventBus.stop()
+    await this.child.close()
+    // @todo: Should we also 'close' the dispatcher?
```
carsonfarmer: No close method on dispatcher, but maybe we could have one? Just so we don't accidentally try to dispatch things on it after we've closed up shop?

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+    return
+  }
+
+  private async onRecord(rec: ThreadRecord) {
```
carsonfarmer: What to do when we get a new (external) thread record...

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (this.threadID?.equals(rec.threadID)) {
+      const logInfo = await this.service.getOwnLog(this.threadID, false)
+      if (logInfo?.id.equals(rec.logID)) {
+        return // Ignore our own events since DB already dispatches to DB reducers
+      }
+      // @todo Should just cache this information, as its unlikely to change
```
carsonfarmer: Yes, that seems like a good idea.

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+        const collection = this.collections.get(value.collection)
+        if (collection !== undefined) {
+          const key = collection.child.prefix.child(new Key(value.id))
+          await this.dispatcher.dispatch({ key, value })
+        }
+      }
+    }
+  }
+
+  private async onEvents(...events: Event[]) {
```
carsonfarmer: What to do when we create a new local event

packages/database/src/db.ts
```diff
@@ -0,0 +1,253 @@
...
+    const id = this.threadID?.bytes()
+    if (id !== undefined) {
+      for (const body of events) {
+        await this.eventBus.push({ id, body })
+      }
+    }
+  }
+
+  private async onUpdate<T extends Entity>(...updates: Update<Op<T>>[]) {
```
carsonfarmer: What to do when that new local event is dispatched and ready to send to listeners

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+  onRetry: err => logger.warn(`create record failed (${err.message}), retrying...`),
+}
+
+export type Events = {
+  record: (rec: ThreadRecord) => void
+}
+
+export type EventJob<T> = { id: Buffer; body: T }
+
+export class EventBus<T = any> extends EventEmitter<Events> {
```
carsonfarmer: The event bus _is_ an event emitter, which is handy. These are typed events, which is even more handy.

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+  private closer?: Closer
+  public queue: Queue<EventJob<T>>
+  constructor(
+    queue: Queue<EventJob<T>> | Datastore<any>,
+    public service: Service,
+    opts: Options = {},
+  ) {
+    super()
+    this.queue = queue instanceof Queue ? queue : new Queue(queue)
+    this.queue.on('next', async (task?: Job<EventJob<T>>) => {
```
carsonfarmer: The queue operates on events as well. I tried to create an asynciterator version, but it felt really hacky and cumbersome, so events it is.

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+      if (task === undefined) return
+      const { job } = task
+      const { id, body } = job
+      const threadID = ThreadID.fromBytes(id)
+      try {
+        await retry(async (_bail, _num) => {
+          // @todo: We could use bail here to bail if the service errors out with a 'headers closed error'
```
carsonfarmer: The headers closed error is what you get when the gRPC proxy server isn't accepting requests (is off). So if we get one of those, we _know_ we're going to probably fail a bunch here. We _could_ take advantage of that information perhaps?

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+          // This would mean that the gRPC service isn't running, i.e., we are in 'offline' mode
+          await this.service.createRecord(threadID, body)
+          // @todo: Add debugging outputs here
+          return this.queue.done()
```
carsonfarmer: Yay, we were successful!

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+        }, merge(retryOpts, opts))
+      } catch (err) {
+        // Skip it for now, we've already retried
+        this.queue.done(true)
```
carsonfarmer: Doh, didn't work... skip this event and move on to the next one (for now)

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+    })
+  }
+
+  private serviceWatcher(id?: ThreadID, start = true) {
```
carsonfarmer: This is pretty simple, subscribe to specific thread events (or all thread events), and rely them to the database (who then relays them to the dispatcher).

packages/database/src/eventbus.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.serviceWatcher(undefined, false)
+    this.queue.stop()
+    await this.queue.close()
+  }
+
+  /**
+   * Push an event onto to the queue
+   * @param event Object to be serialized and pushed to queue via JSON.stringify().
+   */
+  push(event: EventJob<T>) {
```
carsonfarmer: Push a local event onto the queue for processing. Will be processed right away if possible.

packages/database/src/query.ts
```diff
@@ -0,0 +1,158 @@
+// Type definitions for MongoDB 3.3
```
carsonfarmer: Types for Mongodb query objects, just took these from definatelytyped.

packages/database/src/queue.spec.ts
```diff
@@ -0,0 +1,453 @@
+import { expect } from 'chai'
+import { MemoryDatastore, Datastore, Key } from 'interface-datastore'
+import LevelDatastore from 'datastore-level'
+import { isBrowser } from 'browser-or-node'
+import sinon from 'sinon'
+import { Queue } from './queue'
+
+const level = require('level')
+
+describe('Queue', () => {
```
carsonfarmer: Queues are tricky, this helped me test those out. Was based on tests from https://github.com/damoclark/node-persistent-queue

packages/database/src/queue.ts
```diff
@@ -0,0 +1,430 @@
+import { ulid } from 'ulid'
```
carsonfarmer: Not sure what to say about this beyond... its an event-based persistent queue that is useful for offline first applications. This is based on a JS version that I found. It works, it is simple in terms of APIs, and that's about all there is to say about that.

packages/database/src/utils.ts
```diff
@@ -0,0 +1,97 @@
+import {
```
carsonfarmer: Gotta have that utils lib!

packages/service-client/src/index.spec.ts
```diff
@@ -17,15 +17,13 @@ const ed25519 = keys.supportedKeys.ed25519
 async function createThread(client: Client) {
   const id = ThreadID.fromRandom(Variant.Raw, 32)
   const threadKey = Key.fromRandom()
-  const info = await client.createThread(id, { threadKey })
-  return info
+  return client.createThread(id, { threadKey })
```
carsonfarmer: Just a few syntax updates that I discovered along the way. Minor nits.

packages/service-client/src/index.ts
```diff
@@ -122,8 +122,7 @@ export class Client implements Service {
     req.setThreadid(id.bytes())
     req.setKeys(keys)
     const res = (await this.unary(API.CreateThread, req)) as pb.ThreadInfoReply.AsObject
-    const info = threadInfoFromProto(res)
-    return info
+    return threadInfoFromProto(res)
```
carsonfarmer: Nits

packages/service/src/service.spec.ts
```diff
@@ -19,15 +19,13 @@ const ed25519 = keys.supportedKeys.ed25519
...
-  return info
+  return client.createThread(id, { threadKey })
 }
 
 function threadAddr(hostAddr: Multiaddr, hostID: PeerId, info: ThreadInfo) {
   const pa = new Multiaddr(`/p2p/${hostID.toB58String()}`)
   const ta = new Multiaddr(`/thread/${info.id.string()}`)
-  const full = hostAddr.encapsulate(pa.encapsulate(ta)) as any
-  return full
+  return hostAddr.encapsulate(pa.encapsulate(ta)) as any
```
carsonfarmer: same here

packages/service/src/service.ts
```diff
@@ -32,7 +32,7 @@ export class Service implements Interface {
    * @param store The store to use for caching keys and log information.
    * @param client A client connected to a remote service peer.
    */
-  constructor(store: LogStore | Datastore, private client: Client) {
+  constructor(store: LogStore | Datastore, readonly client: Client) {
```
carsonfarmer: We needed this outside the service layer for the database components to use.

packages/service/src/service.ts
```diff
@@ -74,7 +74,21 @@ export class Service implements Interface {
    * @param opts The set of keys to use when adding the Thread.
    */
   async addThread(addr: Multiaddr, opts: KeyOptions) {
-    return this.client.addThread(addr, opts)
+    const logInfo = await this.deriveLogKeys(opts.logKey)
```
carsonfarmer: Improvement here that wasn't needed before we had the db to use. This essentially keeps any core read-level keys on the client _only_, but provides a nice way to cache them.

packages/store/src/stores/store.ts
```diff
@@ -39,6 +39,8 @@ const asUpdate = <T = any>(event: Result<Event<T>>): Update<T> => {
     id: value.id,
     collection: value.collection,
     event: value.patch,
+    // todo: This is a bit of a hack, find a nicer way to do type
```
carsonfarmer: It is indeed... but I didn't want to force a particular type here just because _we're_ using type in our events... But there's a nicer way to handle this with generics, I just haven't done it yet.

packages/database/src/collection.ts
```diff
@@ -0,0 +1,326 @@
...
+ */
+export interface Options<T extends Entity> {
+  child?: Datastore<T>
+  dispatcher?: Dispatcher
+}
+
+/**
+ * Document is a wrapper around a collection and a (proxy to) an entity object.
+ */
+export class Document<T extends Entity = any> {
```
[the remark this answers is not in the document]
sanderpick: Cool!

packages/database/src/collection.ts
sanderpick: Very nice!

packages/database/src/db.spec.ts
sanderpick: Hmm, dunno... hard to weigh in without playing with it. I'd say opt for more JS style!

packages/database/src/eventbus.ts
sanderpick: Does this bubble up to the user?

packages/database/src/eventbus.ts
sanderpick: Do failures make it up to the user somehow? I think this is the reason the queue is just in-mem in go, and can apply back-pressure... the error gets pushed back to the user.

packages/database/src/utils.ts
```diff
@@ -0,0 +1,97 @@
...
+  const pa = new Multiaddr(`/p2p/${hostID}`)
+  const ta = new Multiaddr(`/thread/${threadID}`)
+  return hostAddr.encapsulate(pa.encapsulate(ta))
+}
+
+export interface CacheOptions {
+  duration?: number
+}
+
+export function Cache(params: CacheOptions = {}) {
```
sanderpick: Just curious, what get's cached here?

packages/database/src/db.ts
sanderpick: Hmm, yeah good point. We could change that... GH issue?

packages/database/src/db.ts
sanderpick: +1 for background pull

sanderpick: We'll probably want to allow collections to be passed in the constructor, right? For the same reason we do in go... since you'd start pulling before the collections are in.

packages/database/src/db.ts
sanderpick: This is like the private `newDB` method in Go... this logic would have to go somewhere, seems like a good place for it to me.

packages/database/src/db.ts
sanderpick: Hmm, interesting. Refresher for me, how is the remote peer originally specified?

packages/database/src/db.ts
sanderpick: Maybe in JS, but in Go the adapter will be closed, preventing dispatch from being reached.

packages/database/src/db.ts
carsonfarmer: Oh yes, good call. I'll create a ticket for this!

packages/database/src/db.ts
carsonfarmer: The remote peer is specified via a simple http(s) endpoint with optional port. So, we do know the address component (or at least, one address), but we don't necessarily know which port is its swarm port or grpc port, only its gprc proxy port.

packages/database/src/db.ts
carsonfarmer: Yeh, that is actually the case here as well. No way to actually _call_ dispatch once closed... so should be fine.

packages/database/src/eventbus.ts
carsonfarmer: Well, we don't really _want_ it to bubble up to the user (tho right now, it'll bubble up on the final failure). We _want_ it to just feel like it worked. And then the next time we're online, push out the updates. Hence the persistent queue.

packages/database/src/eventbus.ts
carsonfarmer: See my response above. To support 'offline first', we actually don't want to fail completely if the push to the network doesn't happen. For Go, failure and back pressure are good, because we _should_ be online. But for a browser or local first app, if we can't push out updates, we still want it to 'feel' like things are working.

packages/database/src/db.spec.ts
carsonfarmer: I think I've landed on making the Db _itself_ an event emitter... which is a quick and easy change. That way, it'll just be something like:

```
db.on('some.event', () => console.log)
```

sanderpick:
Great, the point about offline-first makes sense... I see the diff now. 👍 

As for the remote peer info, we should just make an additional peer info API.

packages/database/README.md
```diff
@@ -0,0 +1,57 @@
+# Threads Database _(threads-database)_
+
+[![Made by Textile](https://img.shields.io/badge/made%20by-Textile-informational.svg?style=popout-square)](https://textile.io)
+[![Chat on Slack](https://img.shields.io/badge/slack-slack.textile.io-informational.svg?style=popout-square)](https://slack.textile.io)
+[![npm (scoped)](https://img.shields.io/npm/v/@textile/threads-database.svg?style=popout-square)](https://www.npmjs.com/package/@textile/threads-database)
+[![docs](https://img.shields.io/badge/docs-master-success.svg?style=popout-square)](https://textileio.github.io/js-threads)
+[![standard-readme compliant](https://img.shields.io/badge/standard--readme-OK-green.svg?style=flat-square)](https://github.com/RichardLitt/standard-readme)
```
andrewxhill: isn't this sort of a cheat since you have a bunch of todo's down there? could be an indicator of a non-helpful badge

packages/database/src/db.spec.ts
```diff
@@ -0,0 +1,314 @@
...
+      const eventBus = new EventBus(new DomainDatastore(datastore, new Key('eventbus')), service)
+      const db = new Database(datastore, { dispatcher, service, eventBus })
+
+      const id = ThreadID.fromRandom(Variant.Raw, 32)
+      await db.open(id)
+
+      await db.newCollectionFromObject<DummyEntity>('dummy', {
+        ID: '',
+        name: '',
+        counter: 0,
```
andrewxhill: Are there any character encoding limitation of these instances?

packages/database/README.md
carsonfarmer: well, I guess? its just copy pasta from the others readmes, but the plan _is_ to be standard-readme compliant... could take it out until then?

andrewxhill: or perhaps move the readme TODO sections to tickets and this badge should fail (i think one or two of them are required)

packages/database/src/db.spec.ts
carsonfarmer: According to the JSON Data Interchange Format (https://tools.ietf.org/html/rfc4627):
> JSON text SHALL be encoded in Unicode.  The default encoding is
   UTF-8.
So any Unicode encoding works, which by default in JS is utf-8. So I think we're pretty safe? We could document for sure.

packages/database/README.md
andrewxhill: but then again... i don't think the badge is super informative to a dev/user here. it's an internal check, so maybe better as a step in the lint ci

carsonfarmer: This badge doesn't actually check anything, its just a hardcoded green one. So I'll just create tickets for each sub-package to fill out the readmes and mention that they aren't compliant yet.

carsonfarmer: the standard-readme site just asks you to put the badge on to create 'visibility'

andrewxhill: oohhhh i see. nbd then

packages/database/src/db.spec.ts
andrewxhill: i was a little curious, what if invalid characters are shared from another peer

andrewxhill: (e.g. a peer from go without the same constraints)

carsonfarmer: A peer from go _will_ have the same constraints, because its all JSON under the hood in our collections, so that's good. So both our JS and Go clients will validate the JSON before doing anything so they shouldn't produce any invalid encoding that way. Having said that, an incoming update from some non-compliant peer _could_ probably cause an issue..? Yeh, it'll just fail on write to db, so the user will get an error saying unable to parse JSON I think.

PR #41, opened 2020-03-27
Rename: service => network & entity => instance

### LINKED ISSUE #30: Change: Document/Entity -> Instance
This change is needed to match with go implementation. It shouldn't affect implementation details or anything beyond aesthetics, but will improve the developer experience to have a unified naming convention. After this change, the naming should be updated in the white paper.

### LINKED ISSUE #35: Rename Service to Network/Net
To match Go versions and naming conventions. Note that with this change, the white paper should also be updated.

carsonfarmer:

## Description

Rename Service to Network, and Entity to Instance throughout. Note that we're leaving Document 'as is' because it is simply a wrapper around a Collection/Instance.

Fixes #35 
Fixes #30 

## Type of change

## How Has This Been Tested?

All existing tests pass, and updates were made to reflect the name changes.

## Checklist:

- [ ] ~I have made corresponding changes to the documentation~
- [ ] ~I have added tests that prove my fix is effective or that my feature works~


packages/database/src/collection.ts
```diff
@@ -263,24 +261,24 @@ export class Collection<T extends Entity = any> extends ReadonlyCollection<T> {
   /**
    * Insert (multiple) new entities.
```
andrewxhill: minor

packages/network/src/network.ts
```diff
@@ -49,7 +49,7 @@ export class Service implements Interface {
    * @param id The Thread id.
```
andrewxhill: ID. Wouldn't have wasted the time of a comment, but it got me wondering if we should use ThreadID as convention when describing this particular ID or if it is the "ID of the Thread". Dunno...

andrewxhill: Nah, it's the ID of the Thread. Just thinking of it in the context of instance ID etc.

carsonfarmer: Well, this is indeed a good question IMO. There are a lot of IDs, and the _interface_ is called ThreadID, and also ThreadKey, and maybe even ThreadAuth soon. So maybe good idea to use those names?

carsonfarmer: I'll leave that decision up to a future PR.

packages/database/src/collection.ts
carsonfarmer:
```suggestion
   * Insert (multiple) new instances.
```

PR #135, opened 2020-05-04
Import js-threads-client into js-threads

carsonfarmer:


carsonfarmer:
Tests and linting all pass, but the commit validator is going to fail because some of the incoming commits are from before we were enforcing that @andrewxhill. Otherwise, its really just 17 file updates, several of which are 'admin' type changes.

Issue #55, opened 2020-04-13
Reduce reliance on libp2p-crypto


carsonfarmer:
This is just a really large package in terms of decencies and bundle size. For the most part, we can get away with things like: https://github.com/paulmillr/noble-ed25519, and https://github.com/crypto-browserify/randombytes for our crypto needs.

PR #136, opened 2020-05-04
chore: setup browser builds

### LINKED ISSUE #62: Include browser build step and cdn links
Now that we have folks building things with these libs, we should include browser builds by default. We can stick to webpack, as this is known to work with js-threads libs nicely already.

carsonfarmer:
Adds build steps for browser builds to be included in build output (dist folder).

Fixes #62


carsonfarmer:
I'm going to consider this a 'first pass' at this. There are a lot of optimizations that need to be done here. In particular, #55 would help a lot in terms of bundle sizes. We do some minimizing here, plus the aliases help with reducing duplicated internal code. But we should be able to get these bundle sizes down to under 3Mbs at the very least.

To explore ways to reduce size, a really awesome and easy to use viz tool can be used:

```
cd packages/core
webpack --config ../../webpack.config.js --json > stats.json
npx webpack-bundle-analyzer stats.json
```

This will create a dynamic visualization that let's you explore package deps. Its quite helpful in finding duplicates and outsized bundle dependencies.

Issue #149, opened 2020-05-06
Support Context API in network and network-client libs


carsonfarmer:
As part of this work, it might be useful to move context into its own sub-package?

PR #160, opened 2020-05-07
Standalone context lib for internal use

### LINKED ISSUE #147: Support ThreadID directly in threads-client
It is a bit burdensome to have to call id.toBytes() each time you want to use a ThreadID. Also, consider allowing this to be undefined in newDB, in which case a new random ID would be generated. Thoughts @andrewxhill?

carsonfarmer:

## Description

This creates a new standalone context package to be used by textile, threads, etc. It also attempts to simplify the general DX for getting started with threads-client, particularly in the context of Hub APIs (that require specific headers).

Fixes #147
Partly addresses #149

## Type of change

## How Has This Been Tested?

All existing tests pass.

## Checklist:


packages/client/src/index.ts
```diff
@@ -34,26 +33,42 @@ export { Query, Where }
...
+   * @param ctx Context object containing web-gRPC headers and settings.
+   */
+  async getToken(identity?: Identity, ctx?: Context) {
     const client = grpc.client<pb.GetTokenRequest, pb.GetTokenReply, APIGetToken>(API.GetToken, {
-      host: this.config.host ?? '',
+      host: this.serviceHost,
+      transport: this.rpcOptions.transport,
+      debug: this.rpcOptions.debug,
     })
+    const ident = identity ?? (await Libp2pCryptoIdentity.fromRandom())
```
carsonfarmer, line 63: This let's us avoid explaining random identities up front. Now, if you don't supply an explicit user/identity here, a random one is used. This will be useful for _basic_ demos, because this is a useful demo. And then the second someone asks about user management or identity management, we can point them to the docs about using custom identities.

packages/client/src/index.ts
```diff
@@ -67,14 +82,16 @@ export class Client {
...
           resolve(token)
         } else {
           reject(new Error(code.toString()))
         }
       })
       const req = new pb.GetTokenRequest()
-      req.setKey(identity.public.toString())
-      client.start(this.config.toJSON())
+      req.setKey(ident.public.toString())
+      const metadata = JSON.parse(JSON.stringify(this.context.withContext(ctx)))
```
carsonfarmer, line 93: This change here means callers don't _have_ to use our Context API, then can just use a 'raw' object and it'll still work. This makes this easier to use outside of Typescript, and fits more closely with many other APIs where the 'config' is just config: any...

packages/client/src/index.ts
```diff
@@ -83,12 +100,15 @@ export class Client {
   /**
    * newDB creates a new store on the remote node.
    * @param dbID the ID of the database
+   * @param ctx Context object containing web-gRPC headers and settings.
    */
-  public async newDB(dbID: Buffer) {
+  public async newDB(dbID?: ThreadID, ctx?: Context) {
```
carsonfarmer, line 105: Switching to just using the ThreadID directly. @andrewxhill I would like your opinion on this. Seems simpler, rather than requiring callers to call toBytes() all the time..?

packages/client/src/index.ts
```diff
@@ -83,12 +100,15 @@ export class Client {
...
+    const id = dbID ?? ThreadID.fromRandom()
     const req = new pb.NewDBRequest()
-    req.setDbid(dbID)
+    req.setDbid(id.toBytes())
     await this.unary(API.NewDB, req)
-    return
+    this.context.withThread(id)
```
carsonfarmer, line 110: _miiight_ need to throw a try catch around this in the future?

packages/client/src/index.ts
```diff
@@ -83,12 +100,15 @@ export class Client {
   /**
    * newDB creates a new store on the remote node.
    * @param dbID the ID of the database
+   * @param ctx Context object containing web-gRPC headers and settings.
    */
-  public async newDB(dbID: Buffer) {
+  public async newDB(dbID?: ThreadID, ctx?: Context) {
+    const id = dbID ?? ThreadID.fromRandom()
```
carsonfarmer, line 106: If you _don't_ supply a new ThreadID, create a new random one. This seems like a reasonable thing to do. @andrewxhill?

packages/client/src/index.ts
```diff
@@ -276,23 +296,27 @@ export class Client {
    * @param dbID the ID of the database
    * @param collectionName The human-readable name of the model to use.
    */
-  public readTransaction(dbID: Buffer, collectionName: string): ReadTransaction {
+  public readTransaction(dbID: ThreadID, collectionName: string): ReadTransaction {
     const client = grpc.client(API.ReadTransaction, {
-      host: this.config.host ?? '',
+      host: this.serviceHost,
```
carsonfarmer, line 301: Even if we didn't do this, the default grpc transport etc would be used. But this feels cleaner.

packages/client/src/index.ts
andrewxhill: lovely

packages/client/src/index.ts
andrewxhill: much better

packages/client/src/index.ts
andrewxhill: wayyyy better. oh man, can't believe we missed that

packages/client/src/index.ts
andrewxhill: what if i use a different Id than what's in the current context?

carsonfarmer: The context will be ignored on that one. Actually, that's a good idea. I'm going to also potentially fallback on the context ID if its in there.

packages/client/src/index.ts
andrewxhill: is there a docstring option for WARNING. because we should warn that it's not easy/possible to later migrate an identity to another. so they should be sure they didn't mean to set it. is all

packages/client/src/index.ts
carsonfarmer: I'll push a change that'll fallback on context when calling newDb, but _not_ when starting a transaction, I think that makes sense.

packages/client/src/index.ts
andrewxhill: i'm still having a hard time following the context requirement totally. and in particular if it might be more helpful to just wrap up the context into something more helpful to the develop that lets them choose one of the two possible contexts more semantically. e.g. ctx.userContext() or something, but again I don't really follow it. see comments here https://github.com/textileio/js-textile/issues/140

packages/client/src/index.ts
carsonfarmer: Not that I know of. There is a '@note' one though, so can use that.

packages/client/src/index.ts
carsonfarmer: Hmmm, ok, we _might_ be able to wrap up the main user API Key steps into a single method...
Proposal:

* So they've created the user key via the CLI. Then in their app...
* const ctx = new Context(... host etc).withUserKey(key.secret)

Under the hood, this does:
* const sig = await createAPISig(key.secret) // Defaults to 1 minute from now
* ctx.withAPIKey(key.key).withAPISig(sig)

We could also have a static method that uses all defaults:
* const ctx = Context.fromUserKey(key.secret, host?, transport?, debug?)

carsonfarmer: _That_, coupled with the other changes to client, might make it so that they don't ever actually 'touch' context at all?

carsonfarmer: The trickiness here is that essentially Context is bridging some things that we need for the hub, which we _don't_ need for a local threadsd. So having it available is useful, but we can try to hide more of it by default.

PR #168, opened 2020-05-11
fix(deps): adds core dep in client to right place

andrewxhill:


packages/client/package.json
```diff
@@ -40,6 +40,7 @@
     "@textile/context": "^0.1.0-rc.0",
     "@textile/multiaddr": "^0.0.4-rc.0",
     "@textile/threads-client-grpc": "^0.1.16",
+    "@textile/threads-core": "^0.1.6-rc.0",
```
carsonfarmer, line 43: Is this actually used at run time? Must have missed something...

andrewxhill: https://github.com/textileio/js-threads/blob/master/packages/client/src/index.ts#L10

carsonfarmer: Yeh, its actually this line (which I did miss): https://github.com/textileio/js-threads/blob/master/packages/client/src/index.ts#L66

carsonfarmer: Good catch!

PR #169, opened 2020-05-11
Fixes some dep compile issues

carsonfarmer:
With the upgrade to mingo 0.3, there were some ES import issues that cropped up due to the way the modules were being imported, this should fix that. It also explicitly defines the types returned from several network layer methods to avoid ambiguous types in external usage.

### DIFF
--- a/packages/database/src/collection.ts
+++ b/packages/database/src/collection.ts
@@ -2,11 +2,13 @@ import { Datastore, Key, MemoryDatastore, Query } from 'interface-datastore'
 import Ajv, { ValidateFunction, ValidationError } from 'ajv'
 import { ulid } from 'ulid'
 import { reduce } from 'streaming-iterables'
-import { Query as MingoQuery } from 'mingo/query'
-import { resolve } from 'mingo/util'
+import { Query as MingoQuery } from 'mingo'
 import { JSONSchema4, JSONSchema6, JSONSchema7 } from 'json-schema'
 import { Dispatcher, Instance, JsonPatchStore } from '@textile/threads-store'
 import { FilterQuery } from './query'
+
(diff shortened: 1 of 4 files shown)


packages/network/package-lock.json
```diff
@@ -4,199 +4,6 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"@consento/sync-randombytes": {
```
carsonfarmer, line 7: For some reason, this keeps happening as part of some back and forth on builds. I think, if always built with lerna, _these_ changes are the correct changes.

packages/database/src/collection.ts
```diff
@@ -2,12 +2,14 @@ import { Datastore, Key, MemoryDatastore, Query } from 'interface-datastore'
 import Ajv, { ValidateFunction, ValidationError } from 'ajv'
 import { ulid } from 'ulid'
 import { reduce } from 'streaming-iterables'
-import { Query as MingoQuery } from 'mingo/query'
-import { resolve } from 'mingo/util'
+import { Query as MingoQuery } from 'mingo'
```
carsonfarmer, line 5: Top level import is wrapped, so imports the commonjs-style import. Using the sub-modules directly as before causes the imports to be called outside an es module, which isn't allowed.

packages/network/src/network.ts
```diff
@@ -74,7 +75,7 @@ export class Network implements Interface {
       logKey: logInfo.pubKey,
       token: opts?.token ?? this.token,
     }
-    const info = await this.client.createThread(id, newOpts)
+    const info: ThreadInfo = await this.client.createThread(id, newOpts)
```
carsonfarmer, line 78: Typescript was trying to implicitly define these types, and this causes issues in some cases.

PR #173, opened 2020-05-12
Adds tostring and fromstring methods

carsonfarmer:
For simpler serialization


carsonfarmer:
Probably needs tests, but those can come in a subsequent PR.

PR #175, opened 2020-05-12
Full update to using Context API for managing keys and metadata

### LINKED ISSUE #149: Support Context API in network and network-client libs

carsonfarmer:

## Description

But update to use Context API throughout all js-threads libraries. We now also have better defaults that reflect real developer usage. This also contains several 'nit' fixes for issues with Context and ThreadID in React Native.

Fixes #149 

## Type of change

## How Has This Been Tested?

All existing tests have been updated and pass. New tests have been added to reflect these changes.

## Checklist:

### DIFF
--- a/packages/context/src/index.ts
+++ b/packages/context/src/index.ts
@@ -196,11 +196,12 @@ export class Context {
     transport: grpc.TransportFactory = grpc.WebsocketTransport(),
     debug = false,
   ) {
+    const newContext = { ...json }
+    newContext['host'] = host
+    newContext['transport'] = transport
+    newContext['debug'] = debug
     const ctx = new Context()
-    ctx._context = json
-    ctx._context['host'] = host
-    ctx._context['transport'] = transport
-    ctx._context['debug'] = debug
+    ctx._context = newContext
     return ctx
   }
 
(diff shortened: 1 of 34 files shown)


packages/context/src/index.ts
```diff
@@ -196,11 +196,12 @@ export class Context {
     transport: grpc.TransportFactory = grpc.WebsocketTransport(),
     debug = false,
   ) {
+    const newContext = { ...json }
```
carsonfarmer, line 199: This could lead to some issues with Typescript linters

packages/context/src/index.ts
```diff
@@ -196,11 +196,12 @@ export class Context {
...
+    newContext['host'] = host
+    newContext['transport'] = transport
```
andrewxhill, line 201: it's a bit of bugger running debug since you can't use it as a param without providing grpc transport too. i.e. Context(host, transport, debug), so people wont be able to just run, Context(debug). they have to do ctx.setDebug(true) later. seems counter. Maybe move debug up front?

carsonfarmer: I think I like that suggestion, but it would require updating a lot of tests because we assume a first arg for host in a lot of places. I'm going to leave this for now, and then create a ticket for later? OK?

andrewxhill: yep

PR #200, opened 2020-05-19
Better error on id mismatch

### LINKED ISSUE #191: Error: Thread id mismatch when trying to use `db.open`
Trying to following the instructions in https://docs.textile.io/threads/introduction/#developer-api

If I do the following,

```
    const threadID: ThreadID = ThreadID.fromRandom()
    const db = new Database()
    await db.open({ threadID })
```

I get the error, `Error: Thread id mismatch`

### LINKED ISSUE #192: TypeError: this.child.open is not a function when trying newCollection
Trying to following the instructions in https://docs.textile.io/threads/introduction/#developer-api

I seem to be able to work around the issue https://github.com/textileio/js-threads/issues/191 by just not using `db.open`. Instead, I just do,

```
    const threadID: ThreadID = ThreadID.fromRandom()
    const db = new Database(threadID)
    db.newCollection('dogs', schema)
```

But now the `newCollection` errors with, 

```
(node:28345) UnhandledPromiseRejectionWarning: TypeError: this.child.open is not a function
    at DomainDatastore.open (/Users/andrewhill/Textile/threads/demo/node_modules/datastore-core/src/keytransform.js:20:23)
    at Database.<anonymous> (/Users/andrewhill/Textile/threads/demo/node_modules/@textile/threads-database/src/db.ts:177:22)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/andrewhill/Textile/threads/demo/node_modules/@textile/threads-database/dist/db.js:14:58)
```

I'm using the same schema from the docs,

```json
const schema: JSONSchema = {
  $schema: 'http://json-schema.org/draft-07/schema#',
  title: 'Person',
  type: 'object',
  properties: {
    _id: { type: 'string' },
    name: { type: 'string' },
    age: {
      type: 'number',
      minimum: 0,
      exclusiveMaximum: 100,
    },
  },
}
```

carsonfarmer:
Fixes #191 
Fixes #192

Includes some additional nits, and tests. I went back and forth on various approaches for this, but this felt like the least over-engineering/keeping the fix simple. So went with that.

### DIFF
--- a/packages/database/src/db.spec.ts
+++ b/packages/database/src/db.spec.ts
@@ -2,6 +2,7 @@
 // 'Hack' to get WebSocket in the global namespace on nodejs
 ;(global as any).WebSocket = require('isomorphic-ws')
 
+import path from 'path'
 import { expect } from 'chai'
 import { Multiaddr, ThreadID } from '@textile/threads-core'
 import LevelDatastore from 'datastore-level'
@@ -11,7 +12,7 @@ import { Key } from 'interface-datastore'
 import { DomainDatastore, Dispatcher, Update, Op } from '@textile/threads-store'
 import { Network, Client } from '@textile/threads-network'
 import { MemoryDatastore } from 'interface-datastore'
-import { Database } from './db'
+import { Database, mismatchError } from './db'
 import { EventBus } from './eventbus'
 import { threadAddr } from './utils'
 
@@ -50,7 +51,7 @@ async function runListenersComplexUseCase(los: string[]) {
   // listener's "stream".
   const i1 = new Collection1({ _id: 'id-i1', name: 'Textile1' })
   await i1.save()
-  await delay(1000)
+  await delay(500)
 
   const events: Update[] = []
   // @todo: Should we wrap this in a 'listen' method instead?
@@ -110,7 +111,7 @@ async function runListenersComplexUseCase(los: string[]) {
 }
 
 describe('Database', () => {
-  describe.skip('end to end test', () => {
+  describe('end to end test', () => {
     it('should allow paired peers to exchange updates', async function () {
       if (isBrowser) return this.skip()
       // @todo This test is probably too slow for CI, but should run just fine locally
@@ -166,7 +167,7 @@ describe('Database', () => {
   })
 
   describe('Persistence', () => {
-    const tmp = './test.db'
+    const tmp = path.join(__dirname, './test.db')
     after(() => {
       level.destroy(tmp, () => {
         return
@@ -314,7 +315,14 @@ describe('Database', () => {
       assertEvents(actions, expected)
     })
   })
+
   describe('Basic', () => {
+    const tmp = path.join(__dirname, './test.db')
+    after(() => {
+      level.destroy(tmp, (_err: Error) => {
+        return
+      })
+    })
     it('should return valid addrs and keys for sharing', async () => {
       const store = new MemoryDatastore()
       const db = new Database(store)
@@ -323,6 +331,32 @@ describe('Database', () => {
       const info = await db.getInfo()
       expect(info?.addrs?.size).to.be.greaterThan(1)
       expect(info?.key).to.not.be.undefined
+      await db.close()
+    })
+
+    it('should automatically open if not yet "opened"', async function () {
+      if (isBrowser) return this.skip()
+      const child = new LevelDatastore(tmp)
+      const db = new Database(child)
+      const Col = await db.newCollectionFromObject('blah', { _id: '' })
+      expect(Col).to.not.be.undefined
+      await db.close()
+    })
+
+    it('should throw if our database and thread id do not match', async function () {
+      const store = new MemoryDatastore()
+      let db = new Database(store)
+      await db.open({ threadID: ThreadID.fromRandom() })
+      await db.close()
+      // Now 'reopen' the database
+      db = new Database(store)
+      try {
+        await db.open({ threadID: ThreadID.fromRandom() })
+        throw new Error('should have throw')
+      } catch (err) {
+        expect(err).to.equal(mismatchError)
+      }
+      await db.close()
     })
   })
 })
--- a/packages/database/src/db.spec.ts
+++ b/packages/database/src/db.spec.ts
@@ -119,7 +119,8 @@ describe('Database', () => {
       // Peer 1: Create db1, register a collection, create and update an instance.
       // @note: No identity is supplied, so a random Ed25519 private key is used by default
       const d1 = new Database(new MemoryDatastore())
-      await d1.open()
+      const ident1 = await Database.randomIdentity()
+      await d1.start(ident1)
       const id1 = d1.threadID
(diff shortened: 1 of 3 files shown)


packages/database/src/db.spec.ts
```diff
@@ -110,7 +111,7 @@ async function runListenersComplexUseCase(los: string[]) {
 }
 
 describe('Database', () => {
-  describe.skip('end to end test', () => {
+  describe('end to end test', () => {
```
carsonfarmer, line 114: Enabling this because it's a good test to have going. Only running in node because in browser it can sometimes be slow.

packages/database/src/db.spec.ts
```diff
@@ -314,7 +315,14 @@ describe('Database', () => {
       assertEvents(actions, expected)
     })
   })
+
   describe('Basic', () => {
+    const tmp = path.join(__dirname, './test.db')
```
carsonfarmer, line 320: better cleanup

packages/database/src/db.spec.ts
```diff
@@ -323,6 +331,32 @@ describe('Database', () => {
       const info = await db.getInfo()
       expect(info?.addrs?.size).to.be.greaterThan(1)
       expect(info?.key).to.not.be.undefined
+      await db.close()
+    })
+
+    it('should automatically open if not yet "opened"', async function () {
```
carsonfarmer, line 337: This was the case before, now we're properly testing it.

packages/database/src/db.spec.ts
```diff
@@ -323,6 +331,32 @@ describe('Database', () => {
...
+      if (isBrowser) return this.skip()
+      const child = new LevelDatastore(tmp)
+      const db = new Database(child)
+      const Col = await db.newCollectionFromObject('blah', { _id: '' })
+      expect(Col).to.not.be.undefined
+      await db.close()
+    })
+
+    it('should throw if our database and thread id do not match', async function () {
```
carsonfarmer, line 346: We want to throw on mismatch, and we're using a sentinel error so developers can catch this in their own code

packages/database/src/db.ts
```diff
@@ -23,30 +23,59 @@ const metaKey = new Key('meta')
...
 const duplicateCollection = new Error('Duplicate collection')
 
-export interface InitOptions {
-  dispatcher?: Dispatcher
-  eventBus?: EventBus
-  network?: Network
-  // An optional identity to use for creating records in the database.
-  identity?: Identity
-}
+export const mismatchError = new Error(
```
carsonfarmer, line 26: Exporting this so callers can 'catch' if it they want.

packages/database/src/db.ts
```diff
@@ -23,30 +23,59 @@ const metaKey = new Key('meta')
...
+  `Input ThreadID does not match database ThreadID.
+  Consider creating new database or use a matching ThreadID.
+  Leave ThreadID blank to use default.`,
+)
 
-export interface StartOptions {
+/**
```
carsonfarmer, line 32: Adds missing docs strings

packages/database/src/db.ts
```diff
@@ -63,33 +92,31 @@ export class Database extends EventEmitter2 {
...
+   * @param datastore The primary datastore or a name for a datastore.
+   * It is used to partition out "sub-domains" for collections.
    * @param options A set of database options.
+   * These are used to control the operation of the underlying database.
    */
-  constructor(
-    datastore: Datastore<any> = new LevelDatastore('threads.db'),
-    options: InitOptions = {},
-  ) {
+  constructor(store: Datastore<any> | string, options: Partial<DatabaseSettings> = {}) {
```
carsonfarmer, line 111: Now supports a datastore _or_ a string. But one _must_ be provided. This makes creating the db location explicit, and easier to manage and retrieve later.

packages/database/src/db.ts
```diff
@@ -63,33 +92,31 @@ export class Database extends EventEmitter2 {
...
       options.dispatcher ?? new Dispatcher(new DomainDatastore(datastore, new Key('dispatcher')))
-    // Only if someone _doesn't_ provide a Network, which they should do in most realistic scenarios.
     if (options.network === undefined) {
-      const client = new Client() // @todo: Maybe we should use different defaults here?
-      this.network = new Network(
-        new DomainDatastore(datastore, new Key('network')),
-        client,
-        options.identity,
-      )
+      const client = new Client()
```
carsonfarmer, line 118: In prep for moving to better defaults on the network side.

packages/database/src/db.ts
```diff
@@ -103,15 +130,15 @@ export class Database extends EventEmitter2 {
...
-   * @param datastore The primary datastore.
    * @param options A set of database options.
    */
   static async fromAddress(
     addr: Multiaddr,
-    datastore: Datastore<any>,
+    datastore: Datastore<any> | string,
     threadKey?: ThreadKey,
-    options: InitOptions = {},
+    options: Partial<DatabaseSettings> & StartOptions = {},
```
carsonfarmer, line 141: This should have always been the case, we need to support options from both init and start.

packages/database/src/db.ts
```diff
@@ -165,21 +192,19 @@ export class Database extends EventEmitter2 {
...
-   * @param options A set of options to configure the setup and usage of the underlying database.
+   * underlying services (event bus, network network, etc).
+   * @param opts A set of options to configure the setup and usage of the underlying database.
    */
-  async open(options: StartOptions = {}) {
-    // If we didn't supply an identity upon init, try to create a random one now.
-    await this.network.getToken(this.network.identity ?? (await Libp2pCryptoIdentity.fromRandom()))
+  async open(opts: StartOptions = {}) {
+    // If no identity here, try to use network layer, if not create new random one.
+    await this.network.getToken(
```
carsonfarmer, line 200: It might be better to just not have any defaults at all, and require the developer to always supply an identity? or even just explicitly specify something like "useDefault"?

packages/database/src/db.spec.ts
andrewxhill: lovely

packages/database/src/db.ts
andrewxhill: seems like it's only shortterm utility to have it default. since then it gives a feeling like the app doesn't need to care about the identity. but in reality, the app should store it away to get it back if the memory is flushed or anything. getDefault could be great, saving me from having to setup any idents but also putting them into my own hands. useDefault has the same issue as just having a default.

carsonfarmer: Ok cool, I agree with this. Proposal:

We make identity required here, and then also provide a static method on Database (`Database.randomIdentity()`) so that they have it handy. The, we 'catch' if identity is missing, and in the error: `new Error('Identity required. You may use Database.randomIdentity() to generate a new one if you do not have an existing Identity to use.')`

carsonfarmer: Note also that in the docs strings, we have some wording about identities and how they should be stored off for future use.

PR #214, opened 2020-05-22
Provides token challenge method for getToken

### LINKED ISSUE #52: Ensure grpc handshake in getToken call is properly closed
See here for original question/todo: https://github.com/textileio/js-threads/pull/51/files?file-filters%5B%5D=.json&file-filters%5B%5D=.md&file-filters%5B%5D=.ts#r403431586

### LINKED ISSUE #208: Multistep version of getToken
Right now, getToken handles a bunch of things for the developer so it's simple. Pass in an identity and get back a token. However, if a developer wants to sandbox their users from the APIs a bit more, they may not want to pass a full private key into whatever process is handling this for their user. 

Instead, a flow like the following would be useful:

`getToken(publicKey)` returns a message that needs to be signed and sent back in the next step.

thinking @carsonfarmer has thought about this a bit already.

carsonfarmer:

## Description

Provides a getTokenChallenge method for use by external callers who do not want to pass a user Identity to the clients. This work then wraps this method into our existing getToken promise-based method to maintain API compatibility to to keep around the simpler dev flow.

I have also made identity non-optional in the case where a caller does pass in an identity. This was discussed in a previous PR for the Database, and is a good idea in general.

Fixes #208 
Fixes #52 

## Type of change

## How Has This Been Tested?

All existing tests pass. No new tests were added, as the functionality is entirely encompassed in the existing tests. Some modifications to _some_ tests were made to reflect the slightly different APIs.

## Checklist:

### DIFF
--- a/packages/client/src/index.spec.ts
+++ b/packages/client/src/index.spec.ts
@@ -63,8 +63,7 @@ describe('Client', function () {
 
   describe('.newDB', () => {
     it('response should succeed', async () => {
-      // Calling getToken without an explicit identity will create a default random PKI pair
-      await client.getToken()
+      await client.getToken(identity)
       await client.newDB(dbID)
     })
   })
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -48,53 +48,80 @@ export class Client {
   }
 
   /**
-   * Obtain a token for interacting with the remote network API.
-   * @param identity The generic identity to use for signing and validation.
-   * @param ctx Context object containing web-gRPC headers and settings.
-   * @note If an identity is not provided, a random PKI identity is used. This might not be what you want!
-   * It is not easy/possible to migrate identities after the fact. Please supply an identity argument if
-   * you wish to persist/retrieve user data later.
+   * Create a random user identity.
    */
-  async getToken(identity?: Identity, ctx?: Context) {
+  static async randomIdentity() {
+    return Libp2pCryptoIdentity.fromRandom()
+  }
+
+  /**
+   * Obtain a token for interacting with the remote API.
+   * @param identity A user identity to use for creating records in the database. A random identity
+   * can be created with `Client.randomIdentity(), however, it is not easy/possible to migrate
+   * identities after the fact. Please store or otherwise persist any identity information if
+   * you wish to retrieve user data later, or use an external identity provider.
+   * @param ctx Context object containing web-gRPC headers and settings.
+   */
+  async getToken(identity: Identity, ctx?: Context) {
+    return this.getTokenChallenge(
+      identity.public.toString(),
+      async (challenge: Buffer) => {
+        return identity.sign(challenge)
+      },
+      ctx,
+    )
+  }
+
+  /**
+   * Obtain a token for interacting with the remote API.
+   * @param publicKey The public key of a user identity to use for creating records in the database.
+   * A random identity can be created with `Client.randomIdentity(), however, it is not
+   * easy/possible to migrate identities after the fact. Please store or otherwise persist any
+   * identity information if you wish to retrieve user data later, or use an external identity
+   * provider.
+   * @param callback A callback function that takes a `challenge` argument and returns a signed
+   * message using the input challenge and the private key associated with `publicKey`.
+   * @param ctx Context object containing web-gRPC headers and settings.
+   * @note `publicKey` must be the corresponding public key of the private key used in `callback`.
+   */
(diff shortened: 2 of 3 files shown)


packages/client/src/index.spec.ts
```diff
@@ -63,8 +63,7 @@ describe('Client', function () {
 
   describe('.newDB', () => {
     it('response should succeed', async () => {
-      // Calling getToken without an explicit identity will create a default random PKI pair
-      await client.getToken()
```
carsonfarmer, line 67: Removed this in favor of "explicit is always better than implicit".

packages/client/src/index.ts
```diff
@@ -48,53 +48,80 @@ export class Client {
   }
 
   /**
-   * Obtain a token for interacting with the remote network API.
-   * @param identity The generic identity to use for signing and validation.
+   * Create a random user identity.
+   */
+  static async randomIdentity() {
```
carsonfarmer, line 53: Added here for convenience. I _could_ probably remove this and reduce our dependence on threads-core? But I have some other work in this area that might make this much less of an issue anyway, so adding for now.

packages/client/src/index.ts
```diff
@@ -48,53 +48,80 @@ export class Client {
...
+   * can be created with `Client.randomIdentity(), however, it is not easy/possible to migrate
+   * identities after the fact. Please store or otherwise persist any identity information if
+   * you wish to retrieve user data later, or use an external identity provider.
    * @param ctx Context object containing web-gRPC headers and settings.
-   * @note If an identity is not provided, a random PKI identity is used. This might not be what you want!
-   * It is not easy/possible to migrate identities after the fact. Please supply an identity argument if
-   * you wish to persist/retrieve user data later.
    */
-  async getToken(identity?: Identity, ctx?: Context) {
+  async getToken(identity: Identity, ctx?: Context) {
```
carsonfarmer, line 65: This method stays largely the same, save for the fact that identity is now required.

packages/client/src/index.ts
```diff
@@ -48,53 +48,80 @@ export class Client {
...
+   * A random identity can be created with `Client.randomIdentity(), however, it is not
+   * easy/possible to migrate identities after the fact. Please store or otherwise persist any
+   * identity information if you wish to retrieve user data later, or use an external identity
+   * provider.
+   * @param callback A callback function that takes a `challenge` argument and returns a signed
+   * message using the input challenge and the private key associated with `publicKey`.
+   * @param ctx Context object containing web-gRPC headers and settings.
+   * @note `publicKey` must be the corresponding public key of the private key used in `callback`.
+   */
+  async getTokenChallenge(
```
carsonfarmer, line 87: This is the new method, and allows a caller to do the signing in a callback, to make it easier to isolate keys from the client. Note that this is still an async method that _returns_ the token as a promise, so the caller doesn't need to manage _that_ part of the flow. They only need to provide a single sync or async method to handle the signing of the challenge.

packages/client/src/index.ts
```diff
@@ -48,53 +48,80 @@ export class Client {
...
-          req.setSignature(sig)
+          req.setSignature(signature)
           client.send(req)
+          client.finishSend()
         } else if (message.hasToken()) {
           token = message.getToken()
         }
       })
-      client.onEnd((code) => {
+      client.onEnd((code: grpc.Code, message: string, _trailers: grpc.Metadata) => {
```
carsonfarmer, line 111: Better error handling, so we now know _why_ this error happened, not just the error code.

packages/client/src/index.ts
```diff
@@ -48,53 +48,80 @@ export class Client {
...
-          } catch (err) {
-            reject(err)
-          }
+          const challenge = Buffer.from(message.getChallenge() as string)
+          const signature = await callback(challenge)
           const req = new pb.GetTokenRequest()
-          req.setSignature(sig)
+          req.setSignature(signature)
           client.send(req)
+          client.finishSend()
```
carsonfarmer, line 106: Now that we have sent along our challenge response, we can tell the service we're done. This should help with hanging sockets.

packages/network-client/src/index.ts
```diff
@@ -104,55 +104,80 @@ export class Client implements Network {
   }
 
   /**
-   * Obtain a token for interacting with the remote network API.
-   * @param identity The generic identity to use for signing and validation.
+   * Create a random user identity.
```
carsonfarmer, line 107: This is all the same flow from the db client.

PR #254, opened 2020-06-04
feat(db): adds new init method for network

andrewxhill:


andrewxhill:
slept on the 'scope' change here and still don't like it. reverting in new per

PR #257, opened 2020-06-04
feat(db): adds new init method for network

andrewxhill:


packages/database/src/db.spec.ts
```diff
@@ -361,5 +362,31 @@ describe('Database', () => {
...
+      const store = new MemoryDatastore()
+
+      // We'll just use a dummy auth here. Hub auth should be tested in @textile/hub
+      const auth: UserAuth = {
+        key: '',
+        sig: '',
+        msg: '',
+        token: '',
+      }
+      const db = Database.withUserAuth(auth, store, {}, 'http://localhost:6007')
```
andrewxhill, line 376: the withUserAuth default is remote. so here just using a local process for tests

andrewxhill: given that these will update... should we start implementing a parallel method wherever this exists as, `refreshUserAuth(auth)`?

andrewxhill: or maybe `updateUserAuth(auth)`

carsonfarmer: We probably should yet, but it really doesn't feel right to have that sort of method on the database...

andrewxhill: what are you thinking?

carsonfarmer: I'm not sure yet :P We could certainly add it in for now... I'm just thinking through separation of concerns...

PR #258, opened 2020-06-04
feat(client): add helper method to use KeyInfo directly

andrewxhill:


packages/client/src/index.ts
```diff
@@ -64,6 +64,15 @@ export class Client {
     return new Client(context)
   }
 
+  /**
+   * Create a new gRPC client instance from a supplied key and secret
+   * @param key The KeyInfo object containing {key: string, secret: string, type: 0}. 0 === Account Key, 1 === User Group Key
+   */
+  static async withUserKey(key: KeyInfo, host = defaultHost, debug = false) {
```
andrewxhill, line 71: I mirrored the name in Context here. It got me thinking that these two names aren't quite right though. 

`withUserKey` perhaps, `withKeySecret`

and then

`KeyInfo` perhaps, `KeySecret` with type becoming an optional defaulting to Account type (since you'd probably rarely/never use the keysecret method against the user group key

andrewxhill: take that defaulting bit back... it's backwards. default to user group key

PR #332, opened 2020-06-18
chore(readme): update readme for the primary docs reader

andrewxhill:


README.md
```diff
@@ -63,27 +63,16 @@ Want something specific? Take a look at our [contributor guide](#contributing) f
...
-npm install --save @textile/threads-network @textile/threads-encoding
-```
-
-Similarly, one can install the local storage layer or full-fledged Database separately:
-
-```shell script
-npm i --save @textile/threads-store @textile/threads-database
-``` 
 
 There are also a number of other recommended packages that will make working with Threads in Typescript/Javascript easier:
```
andrewxhill, line 78: which of the below (interface-datastore datastore-level buffer) are still needed?

carsonfarmer: Neither actually, they should come with it now

PR #333, opened 2020-06-22
Adds ability to add from dbinfo

### LINKED ISSUE #342: Schema values being sent 'raw' to gRPC service
When calling `newDBFromAddr`, if a schema is specified in the collections argument, this was being send within the gRPC proto message as a 'raw' object, rather than encoded as a Buffer of a string.

### LINKED ISSUE #343: Add ability to "join" DB from dbInfo object/information.
This would simplify how users can 'join' existing Threads/DBs. Additionally, this should take some of the guesswork out of joining based off 'local' vs 'public' peer addresses. Related to https://github.com/textileio/go-threads/issues/376. This work is slightly more nuanced than it sounds because we need to ensure that a client peer operating on a remote daemon can still add replicators and remote threads, _even if that remote thread already exists on the same remote daemon peer that they are interacting with_.

carsonfarmer:
Implements `joinFromInfo` method that allows a client peer to join a DB from an "invite" with minimal effort. This required some updates on the Go side, namely: https://github.com/textileio/go-threads/pull/378 and https://github.com/textileio/go-threads/pull/380. This also provides the ability to filter out addresses that appear to be 'local' addresses. Note, assuming this API is useful/nice, this should also be implemented for the `database` sub package.

Fixes #343

This PR also includes an unanticipated bug fix where pre-defined Collection schemas were being passed to the gRPC service as 'raw' objects, rather than encoded as a Buffer of a JSON string.

Fixes #342


andrewxhill:
Looks good to me! Would it be worth renaming the KeyInfo type to avoid confusion there, or is it clear enough?

carsonfarmer:
Hmm, Yeh I missed that one. It think the naming change is likely a good one. Though perhaps as a separate PR/fix?

PR #344, opened 2020-06-23
Update ipld block api and reduce crypto deps further

carsonfarmer:
This updates to the latest @ipld/block release, which _drastically_ reduces our crypto dependencies. It also hardcodes `dag-cbor` as our primary encoding format in favor of simpler API calls. Additionally, it updates our webpack configs to reduce our bundle build sizes and times _significantly_.


carsonfarmer:
Not yet sure how to fix that compilation issue with `typescript-docs-verifier`, something to do with how that lib is configured?

carsonfarmer:
Superseded by #404

PR #353, opened 2020-06-29
Allow adding new db from db info + rename

carsonfarmer:


carsonfarmer:
This is a breaking API change, so we'll want to bump the major version.

PR #368, opened 2020-07-02
fix(examples): userauth example doesnt compile

andrewxhill:
since these are compiled into the hub docs, they need to be build-safe.

https://github.com/textileio/js-hub/pull/294/checks?check_run_id=829210901#step:10:203

this fixes that. i can work on adding a build check for examples in this repo too later.


andrewxhill:
Oh... you know. I missed that you did that. Don't see why it shouldn't have... let's see. 

Well... I think in that case you'd not use the callback method. So let me update this a bit.

carsonfarmer:
Yeh its a contrived example for sure.

Issue #381, opened 2020-07-07
Export DBInfo type


andrewxhill:
getDBInfo returns a type that doesn't appear to be accessible.

PR #382, opened 2020-07-07
feat(api): exports dbinfo type for users

andrewxhill:
fixes https://github.com/textileio/js-threads/issues/381


andrewxhill:
turns out this exists, just need to import it from client, threads, or hub library

PR #388, opened 2020-07-10
fix: don't error on listen for deletes fixes #141

### LINKED ISSUE #141: Listening to a Delete event results in a JSON parsing error
When issuing a `client.delete()` on an item, the item is successfully deleted but the `client.listen` handler results in an error:

```
ERROR SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at onMessage (index.js:324)
    at Array.<anonymous> (grpc-web-client.umd.js:1)
```

This is due to the listen method implementation ([ref](https://github.com/textileio/js-threads-client/blob/5aabe1150d9921a3c623460a4b43f64b70162810/src/index.ts#L346)) trying to parse the instance (`Buffer.from(rec.getInstance_asU8()).toString()`) as JSON, but that string is in fact empty string in that case.

I can see how checking the instance string before attempting to parse it could circumvent the error, but it's not clear what to do in that case - how can the callback be invoked to advertise that an instance was deleted? One option would be a third callback that would receive just IDs of deleted instances. Another would be to invoke the first callback with richer metadata about the event, something more akin to the entire `ListenReply` [proto message](https://github.com/textileio/go-threads/blob/e1b3ffe98cb4c525c3ce3de99a4ba26f9e926223/api/pb/api.proto#L184), although with the instance content parsed.

Happy to try and contribute a patch if you suggest a preferred option!

carsonfarmer:

## Description

This checks returned instances _before_ parsing JSON. Also adds a regression test to ensure we don't end up doing this again in the future.

Fixes #141 

## Type of change

## How Has This Been Tested?

All exists tests continue to pass, and a new regression test was added.

## Checklist:

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -741,14 +741,18 @@ export class Client {
       req.addFilters(requestFilter)
     }
 
+    const decoder = new TextDecoder()
+
     const client = grpc.client<pb.ListenRequest, pb.ListenReply, APIListen>(API.Listen, {
       host: this.serviceHost,
       transport: this.rpcOptions.transport,
       debug: this.rpcOptions.debug,
     })
     client.onMessage((message: pb.ListenReply) => {
-      const ret: Instance<T> = {
-        instance: JSON.parse(Buffer.from(message.getInstance_asU8()).toString()),
+      const str = decoder.decode(message.getInstance_asU8())
+      let ret: Instance<T> | undefined
+      if (str !== '') {
+        ret = { instance: JSON.parse(str) }
       }
       nextTick(() => callback(ret))
     })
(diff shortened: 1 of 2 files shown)


packages/client/src/index.ts
```diff
@@ -741,14 +741,18 @@ export class Client {
...
+
     const client = grpc.client<pb.ListenRequest, pb.ListenReply, APIListen>(API.Listen, {
       host: this.serviceHost,
       transport: this.rpcOptions.transport,
       debug: this.rpcOptions.debug,
     })
     client.onMessage((message: pb.ListenReply) => {
-      const ret: Instance<T> = {
-        instance: JSON.parse(Buffer.from(message.getInstance_asU8()).toString()),
+      const str = decoder.decode(message.getInstance_asU8())
```
carsonfarmer, line 752: Note, as is possible, I'll be slowly converting us from requiring Buffer to do this stuff, to using the default TextDecoder. This is a much cleaner, and faster approach.

PR #390, opened 2020-07-13
Remove default browser builds

### LINKED ISSUE #367: Remove support for internal browser builds
This is too much of a moving target, isn't going to cover the use cases of everyone who needs them, and is generally better left to whatever framework someone ends up using threads within. If a standalone build is required, we can ask users/developers to create a ticket for this. We could also consider providing instructions or a default web pack config to use?

carsonfarmer:

## Description

Removes our default browser building steps in favor of using a framework or more custom setups. These browser builds are too difficult to maintain, they slow down our build and testing process, and are almost always not _quite_ what a consumer needs. Additionally, it doesn't really make sense to build the libraries individually, when they are almost always consumed via @textile/threads or @textile/hub. Ideally, users will a) bundle their own browser builds, or b) use a framework such as React or Angular, in which case having these browser builds is not useful anyway. If it comes up, we can provide builds for specific modules. Additionally, I have update our readme with a suggested `webpack.config.js` file.

Fixes #367 

## Type of change

## How Has This Been Tested?

Does not affect internal tests. Browser build tests removed.

## Checklist:


.github/workflows/test.yml
```diff
@@ -34,47 +33,48 @@ jobs:
...
-    - name: Compile Node
-      run: npm run build:node
+      - name: Compile Node
+        run: npm run build
 
-    - name: Compile Browser
-      run: npm run build:browser
+      - name: Compile Test Browser
+        run: |
+          cd packages/threads
```
carsonfarmer, line 76: still building the browser bundle on threads just to make sure it works upon each pr

README.md
```diff
@@ -43,28 +43,10 @@ If you think you've found a bug in Threads, please file a Github issue. Take a l
 
 Textile's Threads Protocol and Database provides an alternative architecture for data on the web. Threads aims to help power a new generation of web technologies by combining a novel use of event sourcing, Interplanetary Linked Data (IPLD), and access control to provide a distributed, scalable, and flexible database solution for decentralized applications. Threads is backed by a great deal of research and experience, and provides protocols for securely storing and sharing content-addressable data (on IPFS), with tooling to help standardize data creation and dissemination.
 
-The primary public API to Threads is the [Threads Database](./packages/database). The Database aims to provide:
```
carsonfarmer, line 46: deleting the cruft

README.md
```diff
@@ -110,12 +92,62 @@ npm run test:node
...
 ```
 
+### Browsers
+
+If you'd like to build browser bundles (say, using webpack), you may want to use something like the following in your `webpack.config.js`. Note however, that if you are working with a framework such as React or Angular, you might be better off letting that bundler handle the tree-shaking etc for you!
+
+```javascript
+const path = require('path')
+
+module.exports = {
```
carsonfarmer, line 108: putting this here for others in case they want it

README.md
```diff
@@ -110,12 +92,62 @@ npm run test:node
...
+      chunks: 'all',
+    },
+    minimize: true,
+  },
+}
+```
+
+The the bundle(s) can be built with:
+
+```bash
```
carsonfarmer, line 144: here's how folks can still do it if they wanna do it manually... which, honestly you shouldn't in 2020

package.json
```diff
@@ -12,9 +12,7 @@
     "bootstrap": "lerna bootstrap",
     "publish": "lerna publish",
     "compile": "lerna run compile",
-    "build": "lerna run build",
-    "build:node": "lerna run build:node --stream",
-    "build:browser": "lerna run build:browser --stream",
+    "build": "lerna run build --stream",
```
carsonfarmer, line 15: only one build script in each lib now

webpack.test.js
```diff
@@ -9,6 +9,9 @@ module.exports = {
       },
     ],
   },
+  performance: {
```
carsonfarmer, line 12: turning this off, since we don't need it. the cool thing is we can actually use this minimal config in our ci script and it works

andrewxhill:
i don't fully understand the implications of this tbh. if you so it's worth it, sgtm!

carsonfarmer:
Essentially, the implications are that we no longer build and release a default browser bundle for each threads sub-package. So in theory, if you wanted to use these sub-packages directly in a html page, you'd have to build the bundle yourself first. _We_ don't ever do this (all demos/examples use webpack or similar), and I think most folks _shouldn't_ do this (you lose out on tree shaking etc). So downstream there aren't really any implications for our libs.

andrewxhill:
should the same be done in the hub? or is it already done the right way?

carsonfarmer:
Well, the hub is doing what I'm removing here, yes: https://github.com/textileio/js-hub/blob/master/packages/buckets/package.json#L16. But this PR is more about _not_ doing it anymore. I think we could also _not_ do it for the Hub APIs. I don't think anyone is using them?

PR #422, opened 2020-07-22
Prefix child db by threadid when starting

carsonfarmer:

## Description

This PR changes the internals of how we "start" Databases. Now, the actual "opening" of the database is deferred until it is actually "started". Additionally, while we can "create" a database and its underlying components (dispatcher, event handler, etc) at construction time, none of this stuff is turned on until we actually start it. Additionally, all DB keys are prefixed by the threadid in a transparent way. So that, in theory, a user doesn't _really_ need to worry about threadid at all. Additionally, this makes the database "name" (which is really just sugar for the developer) default to "thread.db". For developer who want to "sandbox" their db from other user dbs, they can specify their own name.

## Type of change

## How Has This Been Tested?

All existing tests should pass. Additionally, some previous behavior that was expected to fail, should no longer fail, so some checks for this were actually removed. With that in mind, some _additional_ tests should be applied here.

## Checklist:

### DIFF
--- a/packages/database/src/db.ts
+++ b/packages/database/src/db.ts
@@ -33,6 +33,8 @@ const metaKey = new Key("meta")
 const schemaKey = metaKey.child(new Key("schema"))
 const duplicateCollection = new Error("Duplicate collection")
 
+const defaultName = "thread.db"
+
 export const mismatchError = new Error(
   "Input ThreadID does not match existing database ThreadID. Consider creating new database or use a matching ThreadID."
 )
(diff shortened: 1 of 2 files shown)


packages/database/src/db.ts
```diff
@@ -375,45 +404,42 @@ export class Database implements DatabaseSettings {
...
+    this.threadID = info.id
+    if (this.threadID === undefined)
+      throw new Error("Unable to obtain thread id")
+    // Replace default child store with store prefixed by thread id
+    this.child = new DomainDatastore(
+      this.child,
+      new Key(this.threadID.toString())
+    )
+    // Open "new" store, which may in fact already exist
+    await this.child.open()
```
carsonfarmer, line 433: We've deferred start until the last minute here, this makes it possible to prefix all our keys since we might not know the thread id until now.

packages/database/src/db.ts
```diff
@@ -375,45 +404,42 @@ export class Database implements DatabaseSettings {
...
+        throw err
       }
     }
+    // Specify thread id, which could replace an existing one if we are "switching" here
+    this.threadID = info.id
+    if (this.threadID === undefined)
+      throw new Error("Unable to obtain thread id")
+    // Replace default child store with store prefixed by thread id
+    this.child = new DomainDatastore(
+      this.child,
```
carsonfarmer, line 429: I don't _think_ this'll be an issue :P

PR #499, opened 2020-08-19
fix(api) new default api host

andrewxhill:


andrewxhill:
covered in here https://github.com/textileio/js-threads/pull/500

Issue #498, opened 2020-08-19
Add missing @example elements to methods in Client


andrewxhill:
Bunch of these methods don't have examples https://github.com/textileio/js-threads/blob/master/packages/client/src/index.ts

Which creates a lot of empty docs https://textileio.github.io/js-hub/docs/hub.client.find

would be awesome to add those during the refactor.

PR #501, opened 2020-08-20
feat(docs): adds example docs for client apis

andrewxhill:
fixes https://github.com/textileio/js-threads/issues/498


andrewxhill:
@carsonfarmer just adding you here for post-review. this is just a docs update, i couldn't leave it without all these so just went ahead and did my best to add them where i could. i think we could still use better examples of the transaction based methods, the listen filter methods, and the query/queryjson methods.

PR #531, opened 2020-09-10
Adds support for write transaction abort

### LINKED ISSUE #529: Expose method to abort remote transactions
Right now, we don't have a clear way to aport transactions in the threads-client, we should be able to do this.

carsonfarmer:

## Description

This adds support for aborting an "inflight" transaction, rather than calling end (which by default commits the transaction). This is needed, because keeping a transaction "open" (not ending it) will actually block and can be a serious issue if multiple clients are interacting with a remote daemon at the same time. Additionally, in cases where updates need to be rolled back, previously there was no way to manage this without just letting the transaction time out.

Fixes #529 

## Type of change

## How Has This Been Tested?

New tests were added, old tests were reconfigured to address the new API(s). Additionally, a test to check for a transaction "blocking" will be added.

## Checklist:

### DIFF
--- a/packages/client/src/models/ReadTransaction.ts
+++ b/packages/client/src/models/ReadTransaction.ts
@@ -18,6 +18,7 @@ import { Transaction } from "./Transaction"
 
 /**
  * ReadTransaction performs a read-only bulk transaction on the underlying store.
+ * {@inheritDoc @textile/threads-client#Transaction}
  */
 export class ReadTransaction extends Transaction<
   ReadTransactionRequest,
--- a/packages/client/src/models/ReadTransaction.ts
+++ b/packages/client/src/models/ReadTransaction.ts
@@ -19,6 +19,55 @@ import { Transaction } from "./Transaction"
 /**
  * ReadTransaction performs a read-only bulk transaction on the underlying store.
  * {@inheritDoc @textile/threads-client#Transaction}
+ * @example
+ * Create a new entry in our collection
+ * ```typescript
+ * import {Client, ThreadID} from '@textile/threads'
+ *
+ * interface Astronaut {
+ *   name: string
+ *   missions: number
+ *   _id: string
+ * }
+ *
+ * async function createBuzz (client: Client, threadID: ThreadID) {
+ *   const buzz: Astronaut = {
+ *     name: 'Buzz',
+ *     missions: 2,
+ *     _id: '',
+ *   }
+ *
+ *   const t = client.writeTransaction(threadID, 'astronauts')
+ *   await t.start()
+ *   await t.create([buzz])
+ *   await t.end() // Commit
+ * }
+ * ```
+ *
+ * @example
+ * Abort an in-flight transaction
+ * ```typescript
+ * import {Client, ThreadID} from '@textile/threads'
+ *
+ * interface Astronaut {
+ *   name: string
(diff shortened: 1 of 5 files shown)


packages/client/src/models/WriteTransaction.ts
```diff
@@ -197,4 +197,19 @@ export class WriteTransaction extends Transaction<
       this.client.send(req)
     })
   }
+
+  /**
+   * abort quits the current transaction and drops all associated updates.
+   */
+  public async abort(): Promise<void> {
```
andrewxhill, line 204: Can you include an `@example` on WriteTransaction ideally that includes the use abort?

I missed this page in the sidebar, https://textileio.github.io/js-hub/docs/hub.writetransaction I'll fix on upstream though.

packages/client/src/models/ReadTransaction.ts
```diff
@@ -18,6 +18,7 @@ import { Transaction } from "./Transaction"
 
 /**
  * ReadTransaction performs a read-only bulk transaction on the underlying store.
+ * {@inheritDoc @textile/threads-client#Transaction}
```
carsonfarmer, line 21: Is this right for subclasses @andrewxhill ?

packages/client/src/models/WriteTransaction.ts
```diff
@@ -21,6 +21,56 @@ import { Transaction } from "./Transaction"
 
 /**
  * WriteTransaction performs a mutating bulk transaction on the underlying store.
+ * {@inheritDoc @textile/threads-client#Transaction}
```
carsonfarmer, line 24: Same deal here? Also, see examples.

packages/client/src/models/WriteTransaction.ts
carsonfarmer: Example added here, which is a copy of the one also provided at the top-level of the class.

carsonfarmer (commented):
Some comments for you @andrewxhill

packages/client/src/models/ReadTransaction.ts
andrewxhill: yeah, but inheretDoc will only inheret the docstring. nothing else. so examples need to be written on the final class. so for WriteTraction, you should do the example directly on https://textileio.github.io/js-hub/docs/hub.writetransaction

packages/client/src/models/WriteTransaction.ts
andrewxhill: yep, nice, this looks good

packages/client/src/models/WriteTransaction.ts
andrewxhill: boom

packages/client/src/models/ReadTransaction.ts
carsonfarmer: right, did that, and I have to add one for readTransaction too I suppose :)

carsonfarmer: done

carsonfarmer:
Ok @andrewxhill, should be ready for a proper review. If we get this in, I might do a quick pre-release, because I need this for our offline-first threaddb stuff.

andrewxhill (approved):

PR #557, opened 2020-09-24
Support ThreadDB WriteValidators and ReadFilters

### LINKED ISSUE #552: Enable write validators and read filters defined on the client
This is already available on the go-clients, so now we should enabled it here for additional testing/usage. Additionally, some external dev teams are looking forward to using this feature as soon as possible.

carsonfarmer:
## Description

This adds the ability to specify a writeValidator and readFilter on newly created collections and updated collections. It allows callers to specify the function as a "pure" JS ES5 function, and then have it extracted and compiled on the remote, to be run in a safe sandbox environment. This also specifies tests for this functionality, and updates the verify method to reflect a more common usage pattern. Other test nits and fixes also included. Should probably be merged _after_ #554 (which this PR branches from).

Fixes #552 

## Type of change

## How Has This Been Tested?

New tests were created, and all existing tests pass locally, and should also pass on latest docker image release.

## Checklist:


packages/client/src/index.spec.ts
```diff
@@ -242,32 +252,134 @@ describe("Client", function () {
...
+    it("response should contain a JSON parsable instancesList", async function () {
       const instances = await client.create(dbID, "Person", [createPerson()])
       expect(instances.length).to.equal(1)
     })
   })
 
-  describe(".verify", () => {
-    // TODO: Once we support specfying write validators, flush out this test more fully
-    it("should be able to verfiy incoming instances before writing/saving", async () => {
+  describe(".readFilter", function () {
```
carsonfarmer, line 262: Here's where the real changes are.

packages/client/src/index.spec.ts
```diff
@@ -242,32 +252,134 @@ describe("Client", function () {
...
+    interface Dog {
+      _id: string
+      name: string
+      comments: any[]
+    }
+    // You can use typescript types to ensure type safety!
+    const readFilter = (_reader: string, instance: Dog) => {
```
carsonfarmer, line 269: Super cool, you can specify a function in typescript, and it'll get transpiled into es5/commonjs code! If you _aren't_ using a transpiler (like typescript or webpack or something), then you need to specify the function is pure JS, or just provide a string version of it).

packages/client/src/index.spec.ts
```diff
@@ -242,32 +252,134 @@ describe("Client", function () {
...
+    })
+  })
+
+  describe(".writeValidator", function () {
+    interface Dog {
+      _id: string
+      name: string
+      comments: any[]
+    }
+    const writeValidator = (_writer: string, event: any, _instance: Dog) => {
```
carsonfarmer, line 294: I'd like to specify a more nuanced type for event, I just haven't added it here yet. I'll do it in another PR.

packages/client/src/index.spec.ts
```diff
@@ -242,32 +252,134 @@ describe("Client", function () {
...
+      try {
+        await client.delete(dbID, "Dog", [id])
+        throw new Error("should have failed")
+      } catch (err) {
+        expect(err.message).to.include("app denied net record body")
+      }
+    })
+  })
+
+  describe(".verify", function () {
```
carsonfarmer, line 339: This has been updated, and mostly matches the go test (so do the above two)

packages/client/src/index.spec.ts
```diff
@@ -277,36 +389,34 @@ describe("Client", function () {
...
       const instances = await client.create(dbID, "Person", [createPerson()])
       expect(instances.length).to.equal(1)
       const personID = instances[0]
       const deleted = await client.delete(dbID, "Person", [personID])
       expect(deleted).to.be.undefined
     })
   })
-  describe(".has", () => {
-    it("the created object should also return true for has", async () => {
+  describe(".has", function () {
```
carsonfarmer, line 401: These are also just because mocha behaves better when you don't use arrow functions

packages/client/src/index.spec.ts
```diff
@@ -317,46 +427,45 @@ describe("Client", function () {
       expect(foundPerson["_id"]).to.equal(personID)
     })
   })
-  describe(".findById", () => {
-    it("response should contain a JSON parsable instance property", async () => {
-      const instances = await client.create(dbID, "Person", [createPerson()])
-      const personID = instances.pop()!
-      const find = await client.findByID<Person>(dbID, "Person", personID)
-      expect(find).to.not.be.undefined
-      expect(find).to.haveOwnProperty("instance")
```
carsonfarmer, line 326: This is important, because it provides breaking changes. We pull out the wrapped instances here. Since we're breaking API for the 1.0 release anyway, this is a _much_ better DX!

packages/client/src/index.ts
```diff
@@ -37,27 +36,71 @@ export {
...
+      /(?:\/\*(?:[\s\S]*?)\*\/)|(?:([\s;])+\/\/(?:.*)$)/gm,
+      "$1"
+    )
+  }
+  const s = removeCommentsFromSource(fn.toString())
+  return s.substring(s.indexOf("{") + 1, s.lastIndexOf("}"))
 }
 
-export interface CollectionConfig {
+/**
```
carsonfarmer, line 62: Put the docs for the validator/filter here... is that a good idea @andrewxhill ?

packages/client/src/index.ts
```diff
@@ -37,27 +36,71 @@ export {
...
+ * The function must return a JavaScript object. Most implementation will modify and return the
+ * current instance.
+ * Like write validation, read filtering opens the door to a variety of app-specific logic.
+ * Textile Buckets file-level access roles are implemented in part with a read filter.
+ */
+export interface CollectionConfig<W = any, R = W> {
   name: string
-  schema: any
-  indexes: pb.Index.AsObject
+  schema?: JSONSchema3or4 | any // Union type to indicate that JSONSchema is preferred but any works
```
carsonfarmer, line 94: I have both here, and even though "any" will catch everything, it is helpful to have this "helper" type to do compile time checking on the validity of your json schema

packages/client/src/index.ts
```diff
@@ -37,27 +36,71 @@ export {
...
+  indexes?: pb.Index.AsObject[]
+  writeValidator?:
```
carsonfarmer, line 96: write validator can be generically typed... which is pretty nice, so you can do type checking on your input function for better safety on the remote!

packages/client/src/index.ts
```diff
@@ -75,7 +118,8 @@ export enum Action {
   DELETE,
 }
 
-export interface Update<T = any> extends Instance<T> {
+export interface Update<T = unknown> {
+  instance: T | undefined
```
carsonfarmer, line 122: Pulls the instance up to the top level

packages/client/src/index.ts
```diff
@@ -122,7 +166,7 @@ export interface DBInfo {
  *     missions: 2,
  *     _id: '',
  *   }
- *   await client.newCollectionFromObject(threadID, 'astronauts', buzz)
+ *   await client.newCollectionFromObject(threadID, buzz, { name: 'astronauts' })
```
carsonfarmer, line 169: note the new call signatures

packages/client/src/index.ts
```diff
@@ -211,7 +253,7 @@ export class Client {
    * @remarks
    * See `PrivateKey`
    */
-  static async randomIdentity(): Promise<PrivateKey> {
+  static randomIdentity(): PrivateKey {
```
carsonfarmer, line 256: this isn't async anymore, which is handy. it won't "break" anything downstream, but it _is_ different/api breakage

packages/client/src/index.ts
```diff
@@ -318,18 +360,19 @@ export class Client {
...
     req.setDbid(dbID.toBytes())
     if (name !== undefined) {
       this.context.withThreadName(name)
       req.setName(name)
     }
-    await this.unary(API.NewDB, req)
-    // Update our context with out new thread id
-    this.context.withThread(dbID.toString())
-    return dbID
+    return this.unary(API.NewDB, req, () => {
```
carsonfarmer, line 371: Better type safety here now, and cleaner call signatures

packages/client/src/index.ts
```diff
@@ -372,59 +421,56 @@ export class Client {
...
+      }
+      return dbs
+    })
   }
 
   /**
    * Lists the collections in a thread
    * @param thread the ID of the database
    */
-  public async listCollections(
```
carsonfarmer, line 406: This makes transpiration cleaner by removing the async declarations... now we just return a promise, which is easier for the transpiler to deal with?

packages/client/src/index.ts
```diff
@@ -450,42 +496,99 @@ export class Client {
    * }
    *
    * async function newCollection (client: Client, threadID: ThreadID) {
-   *   return await client.updateCollection(threadID, 'astronauts', astronauts)
+   *   return await client.updateCollection(threadID, { name: 'astronauts', schema: astronauts })
+   * }
+   * ```
+   * @example
```
carsonfarmer, line 502: docs!

packages/client/src/index.ts
```diff
@@ -618,35 +721,42 @@ export class Client {
...
     req.setName(name)
-    const res = (await this.unary(
+    return this.unary(
       API.GetCollectionInfo,
-      req
-    )) as pb.GetCollectionInfoReply.AsObject
-    res.schema = JSON.parse(
-      (Buffer.from(res.schema as string, "base64") as unknown) as string
+      req,
+      (res: pb.GetCollectionInfoReply) => {
```
carsonfarmer, line 748: This is just a much cleaner response

packages/client/src/index.ts
```diff
@@ -618,35 +721,42 @@ export class Client {
...
+        const result: CollectionConfig = {
+          schema: JSON.parse(decoder.decode(res.getSchema_asU8())),
+          name: res.getName(),
+          indexes: res.getIndexesList().map((index) => index.toObject()),
+          // We'll always return strings in this case for safety reasons
```
carsonfarmer, line 753: Because we don't wanna "parse" arbitrary functions from a remote system, in case someone messes with us!

carsonfarmer: This "parsing" is on the bytes, so I think we're safe, because there's no "eval" going on anywhere in the stack.

packages/client/src/index.ts
```diff
@@ -680,17 +790,28 @@ export class Client {
...
-    // Hacky way to extract threadid from addr that succeeded
-    // @todo: Return this directly from the gRPC API?
-    const result = new Multiaddr(Buffer.from(req.getAddr_asU8()))
-      .stringTuples()
-      .filter(([key]) => key === 406)
-    return ThreadID.fromString(result[0][1])
+    return this.unary(API.NewDBFromAddr, req, (/** res: pb.NewDBReply */) => {
+      // Hacky way to extract threadid from addr that succeeded
+      // TODO: Return this directly from the gRPC API on the go side?
+      const result = new Multiaddr(Buffer.from(req.getAddr_asU8()))
```
carsonfarmer, line 810: This todo is a good one, I should create a ticket in go-threads for this... would be really handy

carsonfarmer (commented):
Some comments for your review pleasure @andrewxhill

packages/client/src/index.ts
andrewxhill: looks right to me, yeah

packages/client/src/index.ts
```diff
@@ -37,27 +36,71 @@ export {
...
+ * * instance: The current instance as a JavaScript object before the update event is applied.
+ *
+ * A falsy return value indicates a failed validation.
+ *
+ * Having access to writer, event, and instance opens the door to a variety of app-specific logic.
+ * Textile Buckets file-level access roles are implemented in part with a write validator.
+ *
+ * The `readFilter` function receives three arguments:
+ * * reader: The multibase-encoded public key identity of the reader.
+ * * instance: The current instance as a JavaScript object.
```
andrewxhill, line 85: not certain how this will render with api-documenter. i think i recall that bullet lists will be smashed together

packages/client/src/index.ts
andrewxhill: we could remove this. seems pretty well covered now just by importing PrivateKey

carsonfarmer: even better! I'll do that 👍

andrewxhill (approved):
cool, some nice simplifications in there

packages/client/src/index.ts
carsonfarmer: ah ok, but the newlines should work right? I can just leave the bullets out

andrewxhill: i take it back https://textileio.github.io/js-hub/docs/hub.buckets.copyauth

carsonfarmer: perfect! in it stays!

andrewxhill: here's the format for that https://github.com/textileio/js-hub/blob/master/packages/grpc-authentication/src/index.ts#L64

PR #604, opened 2020-11-12
feat(bundle): try out bili for packaging

andrewxhill:


packages/client/package.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "@textile/threads-client",
   "version": "1.1.0",
-  "main": "dist/index",
-  "types": "dist/index",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "module": "dist/index.esm.ts",
```
andrewxhill, line 6: bili now doing all of them in one go. not using lerna build here, though we can probably back it into that

carsonfarmer:
closing this as #618 now supersedes this. but let's keep bili on our radar.

PR #633, opened 2020-12-09
Single file modules

### LINKED ISSUE #62: Include browser build step and cdn links

### LINKED ISSUE #575: Add vanilla JS examples
I’ve been watching this project for over a year, waiting for vanilla JS examples.

I’m always frustrated when all the examples are implemented in React.

carsonfarmer:
## Description

This PR adds "single-file" ES6 module builds for all our sub-packages. In actual fact, they aren't quite single-file modules. They expect to be able to import @textile/* modules where possible. But otherwise, bundle up their required "vendor" deps. This is a pretty nice pattern for using ES6 modules in browsers that support remote importing of module deps. For instance, we can do things like:

```
import("https://unpkg.com/@textile/threads-client?module")
```

and it should automatically try to import other @textile/* libs that support ES6 module builds.

**note**: You can ignore most files for the review, I have included comments to help guide your review.

Fixes #62
Fixes #575 and others

Some BREAKING CHANGES

* ES module builds are now single file with imports. This shouldn't actually affect any real-world usage, but it is a breaking change, so we should bump the major versions.
* Errors have been moved from top-level exports to "namespaced" exports in the security and context submodules. This WILL affect downstream tests (in `js-textile/packages/users/src/users.spec.ts` and `js-textile/packages/hub-threads-client/src/client.spec.ts`).
* The build process has been simplified and made faster. But, it now requires a slightly more "manual" process for releases! We have removed our prereleaseOnly build script in favor of doing in manually. This is because the rollup build is top-level across all lerna packages, so we don't want each sub-package to run its own local build script. That means that the publish process is now: `npm run publish`, and the script does the rest for us.

## Type of change

## How Has This Been Tested?

All existing tests continue to pass. New tests should be created to ensure the browser builds are usable directly in a browser. This has not been done yet.

## Checklist:


package.json
```diff
@@ -10,12 +10,12 @@
   "scripts": {
     "clean": "lerna clean && lerna run clean",
     "bootstrap": "lerna bootstrap",
-    "publish": "lerna publish",
-    "compile": "lerna run compile",
-    "build": "lerna run build --stream",
-    "commitmsg": "commitlint -e $GIT_PARAMS",
+    "publish": "npm run build && lerna publish",
```
carsonfarmer, line 13: This makes for a slightly more "manual" build process, but it is significantly faster, and does more things at the top level. It will also now do all the sub-package builds and type-checking in parallel as much as it can.

packages/client/CHANGELOG.md
```diff
@@ -3,6 +3,87 @@
 All notable changes to this project will be documented in this file.
```
carsonfarmer, line 3: These are all automatically generated, but since we're working in a branch, they show up here. You can ignore these.

packages/client/src/index.ts
```diff
@@ -23,6 +23,7 @@ import {
 import { ThreadID } from "@textile/threads-id"
 import toJsonSchema, { JSONSchema3or4 } from "to-json-schema"
 import {
+  ComparisonJSON,
```
carsonfarmer, line 26: Exporting these for an external dev user

packages/multiaddr/src/codec.ts
```diff
@@ -5,9 +5,6 @@ import varint from "varint"
 import * as convert from "./convert"
 import { protocols } from "./protocols"
 
-const uint8ArrayConcat = require("uint8arrays/concat")
```
carsonfarmer, line 8: Was able to pull these out and use native functionality instead. This saves us bundle size, and also better builds for ES modules.

packages/multiaddr/src/convert.ts
```diff
@@ -1,10 +1,11 @@
 import { ThreadID } from "@textile/threads-id"
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
+import Convert from "multiaddr/src/convert"
```
carsonfarmer, line 4: Avoiding things like "require" here is useful, because now we can use auto-encoders to build our ES modules. Otherwise, we have to dance around require statements etc. The ts-ignore is not ideal, but it allows us to import from paths like that, and we don't really need the type checking here anyway, because we're just passing our args down to that internal function anyway.

packages/security/src/index.ts
```diff
@@ -142,12 +142,4 @@ export async function createUserAuth(
   }
 }
 
-/**
```
carsonfarmer, line 145: In previous PR (which was rebased upon here), this was moved to context

packages/security/src/key.ts
```diff
@@ -1,7 +1,9 @@
 import randombytes from "@consento/sync-randombytes"
 import multibase from "multibase"
 
-export const invalidKeyError = new Error("Invalid key")
+export const errors = {
```
carsonfarmer, line 4: Keep our errors in "namespaces" to avoid them ending up in the top level export in ES modules.

packages/threads/package.json
```diff
@@ -13,24 +13,21 @@
...
   "repository": "github:textileio/js-threads",
   "devDependencies": {
     "@improbable-eng/grpc-web": "^0.13.0"
   },
   "dependencies": {
-    "@textile/crypto": "^0.1.2",
-    "@textile/security": "^0.6.2",
-    "@textile/threads-client": "^1.3.2",
-    "@textile/threads-id": "^0.3.1",
-    "fast-sha256": "^1.3.0"
```
carsonfarmer, line 31: dropped some deps and bumped others!

rollup.config.js
```diff
@@ -0,0 +1,84 @@
+import { getPackages } from "@lerna/project"
```
carsonfarmer, line 1: here's the real new stuff. we're using rollup here to create single-file ES modules where all @textile/ libs are treated as "external", but any other "vendor" libs are bundled within the single file module. This is awesome for things like importing from unpakg and similar that support ES modules. Additionally, for those running ES modules themselves locally, they can simple npm install any textile deps and then serve up the whole dir for their ES modules. I think its a pretty nice release pattern.

carsonfarmer: this is actually tightly integrated with lerna, so it actually uses lerna and rollup together to automatically rollup each sub package. we do this at the top-level so we can take advantage of lerna's build system for correctly ordering builds, while also using rollups awesome bundler features.

rollup.config.js
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * @returns Promise<RollupConfig>
+ */
+async function main() {
+  const config = []
+  // Support --scope and --ignore globs if passed in via commandline
+  const { scope, ignore } = minimist(process.argv.slice(2))
+  const packages = await getSortedPackages(scope, ignore)
+  packages.forEach((pkg) => {
```
carsonfarmer, line 49: the actual release targets are defined in each sub-package's package.json file. so its pretty easy to adapt to different targets. for instance, we could pretty easily add builds for older browsers, or umds etc.

rollup.config.js
```diff
@@ -0,0 +1,84 @@
...
+      output: [
+        {
+          exports: "named",
+          file: path.join(basePath, "dist/esm/index.js"),
+          format: "es",
+        },
+      ],
+      external,
+      plugins: [
+        json(),
```
carsonfarmer, line 66: we support json imports, resolving npm modules, common js deps (which get bundled into the es module) and typescript. but we don't actually type check here, we leave that up to the regular build step

carsonfarmer (commented):
Here are some guiding comments for you @andrewxhill.

packages/multiaddr/src/convert.ts
asutula: I don't really understand most of anything in this pr because this is pretty exotic stuff, but just curious, what exactly are you importing here? seems like you're referencing this same file?

packages/security/src/key.ts
asutula: so "namespace" is just an object you're exporting. slick idea.

asutula (approved):
Cool stuff, but I actually don't understand much of it. I'd like to. Merge away, and it might be cool to get a brain dump from you on the state of the art for js lib building and packaging. I've been keeping powergate client stuff intensionally very simple, but I'm sure we can benefit from all your new knowledge there.

packages/security/src/key.ts
carsonfarmer: Yeh. In ES modules world (and TS really), the "best" way forward would be to have an "errors.ts" module, and export from that. And then do things like `import { Error } from "./errors"`, but that is overkill for something like this. There is a "namespace" pattern in TS, but it is mostly deprecated in favor of doing the above or the "object" pattern.

packages/multiaddr/src/convert.ts
carsonfarmer: Well, in this case, that is just unfortunate naming ;P I'm actually importing from the external multiaddr lib here (note the absolute vs relative path). I'm doing it this way because we have "hack" our own threads protocol into that lib, but that lib (unlike the go version) doesn't actually provide a way to register new protocols. But I don't want to just fork or otherwise copy-pasta the whole lib here, so I'm importing their internal Convert module and using it directly here. Its defs a hack... but one that is pretty common in JS land (but not TS land).

carsonfarmer:
Note to self: Once we merge this, we'll need to "upgrade" the lerna pre-release packages to full releases. This will also bump the dist tag from "next" to "latest".

